PR #20, opened 2024-06-09
Feature/axi master full

obruendl:
* ADD: Add *olo_axi_master_full*
* BUGFIX: Reduce UserTransactionSizeBits_g default of *olo_axi_master_simple* to 24 (32 as before was illegal)
  * Regarded backward compatible - because any instantiation using the default would fail anyways
* IMPROVEMENT: Added more assertions to *olo_axi_master_simple*
* IMPROVEMENT: Cover InWidth=OutWidth for *olo_base_wconv_n2xn*


obruendl:
Replaced by #21

Issue #43, opened 2024-07-24
VSG Linting


obruendl:
@patrick-studer implements VSG linting.

The issue is meant to discuss linting and linting-rule related questions

patrick-studer:
## FSM- and Enum Types

### Observation

#### FSM
- \src\axi\vhdl\olo_axi_lite_slave.vhd (1 hit)
  - `type Fsm_t is (Idle, WrCmd, WrData, WrResp, RdCmd, RdData, RdResp);`
- \src\axi\vhdl\olo_axi_master_full.vhd (5 hits)
  - `type WriteCmdFsm_t is (Idle_s, Apply_s);`
  - `type WriteWconvFsm_t is (Idle_s, Transfer_s);`
  - `type WriteAlignFsm_t is (Idle_s, Transfer_s, Last_s);`
  - `type ReadCmdFsm_t is (Idle_s, Apply_s, WaitDataFsm_s);`
  - `type ReadDataFsm_t is (Idle_s, Transfer_s, Wait_s);`
- \src\axi\vhdl\olo_axi_master_simple.vhd (5 hits)
  - `type WriteTfGen_s is (Idle_s, MaxCalc_s, GenTf_s, WriteTf_s);`
  - `type ReadTfGen_s is (Idle_s, MaxCalc_s, GenTf_s, WriteTf_s);`
  - `type AwFsm_s is (Idle_s, Wait_s);`
  - `type ArFsm_s is (Idle_s, Wait_s);`
  - `type WFsm_s is (Idle_s, NonLast_s, Last_s);`
- \src\intf\vhdl\olo_intf_i2c_master.vhd (1 hit)
  - `type Fsm_t is ( BusIdle_s, BusBusy_s, MinIdle_s, Start1_s, Start2_s, WaitCmd_s, WaitLowCenter_s,`
- \src\intf\vhdl\olo_intf_spi_master.vhd (1 hit)
  - `type State_t is (Idle_s, SftComp_s, ClkInact_s, ClkAct_s, CsHigh_s);`

#### ENUM
- none

#### ARRAY
- \src\base\vhdl\olo_base_arb_prio.vhd
  - type `Data_t is array (natural range<>) of std_logic_vector(Out_Grant'range);`
- \src\base\vhdl\olo_base_cc_pulse.vhd
  - type `Pulse_t is array (natural range <>) of std_logic_vector(NumPulses_g - 1 downto 0);`
- \src\base\vhdl\olo_base_delay.vhd
[...]

PR #44, opened 2024-07-24
Feature/linting

obruendl:
Add VSG Linting

#43


obruendl:
@patrick-studer 

I changed the python setup in the YAML files (see https://github.com/open-logic/open-logic/pull/45). The advantage is that the pip state now is preserved (shorter pipeline runtimes because no python installations are required).

To add python packages, just add them to .github/workflows/requirements.txt

obruendl:
To add apt packages, now use To add python packages, just add them to .github/workflows/apt-dependencies.txt

I updated the YAML structure a bit, so all jobs can use the same runner setup (and all get same python, same apt, same tools ,etc). My YAML rework is done - should stay stable now.

obruendl:
I accidentally closed the PR because some errors I did when merging develop to master for the 2.1.0 release. I now reopened it and resolved the conflicts that emerged.

@patrick-studer The one thing you need to fix: Install VSG using the requirements.txt file (instead of through the install script) because the CI/CD setup was changed.

obruendl (commented):
Reviewed for reverse compatibility

PR #45, opened 2024-07-25
Feature/vunit junit report

obruendl:
Add XUnit report to VUnit
Use python-setup action and cache (instead of custom script)


obruendl (commented):
All fine

PR #49, opened 2024-08-02
Allow simulators other than GHDL and ModelSim in run.py

nickg:
If the `VUNIT_SIMULATOR` environment variable is already set then use that instead of overriding it.  This allows the regression tests to run with the VHDL simulator I am developing at [nickg/nvc](https://github.com/nickg/nvc).

As far as I know it's fine to set simulator-specific VUnit options (e.g. `ghdl.elab_flags`) when that simulator is not selected.

### DIFF
--- a/sim/run.py
+++ b/sim/run.py
@@ -18,10 +18,11 @@ if "--coverage" in sys.argv:
         "Coverage is only allowed with --modelsim"
 
 # Obviously the simulator must be chosen before sources are added
-if USE_GHDL:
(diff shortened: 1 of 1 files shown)


obruendl:
@nickg 

Thank you for your contribution. 

Before merging I will install NVC (I am not used to it yet) and try it out. I would want to be able testing it on my side for maintenance reasons.

However, I am on my way to vacation - so you should expect a delay of 2 weeks or so. I hope this is not a problem to you.

PR #89, opened 2024-11-04
Fix typo in README.md

Monish-VR:
In the starting phrase of readme "Open Logic aims to be for HDL projects what what stdlib is for C/C++ projects." , the word 'what' repeats two times which may confuse the readers.


obruendl (approved):
Obvious. Thanks.

Issue #109, opened 2024-12-22
[Fusesoc] Redundant provider key in core files


rbrglez:
## 1. Summary

When adding the `open-logic` repository to a `fusesoc` project, the entire repository is downloaded, but `fusesoc core list` displays the cores as empty, even though the source code for these cores is already present in the project directory. Additionally, running the project causes `fusesoc` to download the `open-logic `cores again into `~/.cache/fusesoc`, even though they are already available locally. The issue is identified to be caused by the presence of the `provider` key in the `open-logic` `*.core` files

Excerpt from [olo_base.core](https://github.com/open-logic/open-logic/blob/main/src/base/olo_base.core):
```yml
CAPI=2:

name : open-logic:open-logic:base:3.0.2
description : ...

filesets:
  rtl:
    files:
      - vhdl/olo_base_cc_n2xn.vhd
      - ...
      - vhdl/olo_base_cc_xn2n.vhd
    file_type : vhdlSource
    logical_name : olo

targets:
  default:
    filesets :
      - rtl
  
provider:
  name    : github
  user    : open-logic
  repo    : open-logic
  version : 3.0.2
```

The core files in the `open-logic` repository, include a `provider` key. This key points to the GitHub repository containing the source code, which is necessary for example the [fusesoc-cores](https://github.com/fusesoc/fusesoc-cores) repository (where only core files are present, not the source code). However, in the context of the current project, this key should not be included because the source code for the `open-logic` cores is already locally available.

## 2. Example Configuration

Adding `open-logic` to my project
```bash
fusesoc library add open-logic https://github.com/open-logic/open-logic
``` 

### 2.1 Configuration with `provider` key

Listing all my cores in project. Note that `open-logic` cores are listed as empty, even though whole repository was downloaded into my project.
```bash
$ fusesoc core list

Available cores:

Core                                         Cache status  Description
================================================================================
::vga_timing_core:0.1.0                     :      local : VGA Timing Core
::xbc_testbench_utils:0.1.0                 :      local : XBC Testbench Utility Package
open-logic:open-logic:axi:3.0.2             :      empty : AXI related modules see https://github.com/open-logic/open-logic/blob/main/doc/EntityList.md#axi
open-logic:open-logic:base:3.0.2            :      empty : Basic Circuitry (e.g. FIFOs, CDCs, ...) see https://github.com/open-logic/open-logic/blob/main/doc/EntityList.md#base
[...]

PR #110, opened 2024-12-27
FEATURE: Allow using local open-logic repo through FuseSoC

obruendl:
reported by @rbrglez

refs #109


rbrglez:
Check out this repository: https://github.com/rbrglez/open-logic-example

Here I have tested inclusion of open-logic library to a project with fusesoc.
- **Option 1**:  using only `*.core` file with provider-key
- **Option 2**: including open-logic as a git submodule to the project and adding it as a local fusesoc library. 


After testing different options of including open-logic with fusesoc I found out that I really dislike having development **and** stable cores in a single branch. I suggest that development branch has only development cores, while release branch (e.g., `main`) has only stable cores.

src/base/olo_base.core
rbrglez, line 1: This is not stable core!

It is true, that it fetches stable open-logic repository to `~/.local/fusesoc`, but It doesn't use fetched repository. Instead it uses development repository!

Reason for this is, because it doesn't find `vhdl/<file_name>.vhd` in `~/.local/fusesoc/open-logic_open-logic_base_3.0.2`, that is why it defaults to using files in current repository which are unstable!

If you would want to have intended behavior you must have correct paths.
`src/base/vhdl/<file_name>.vhd`

I think I should report this behavior to fusesoc as a bug

Check out this repository, to verify this behavior (check out README)
https://github.com/rbrglez/open-logic-example

obruendl:
Whats the exact reason for the dislike? Not speaking against it - just trying to understand downsides of the approach.

rbrglez:
> Whats the exact reason for the dislike? Not speaking against it - just trying to understand downsides of the approach.

After I have tested including multiple open-logic versions to my project with fusesoc I have found out that inclusion of stable cores in development version is a bit redundant. If I would want to have both development and stable cores in my project I want to have control which ones to add.

Play a bit around in example project and tell me what do you think

obruendl:
You should be able to control which ones to include (add dependency to) by either using the "open-logic-dev" or the "open-logic" library. At least this is my understanding.

Us there any problem with this?

rbrglez:
Yes, this would work fine (if release core would have correct paths https://github.com/open-logic/open-logic/pull/110#discussion_r1898671210)

It is just my preference. When I add development open-logic library to my project I want to have only development cores. If I would like to have also stable cores then I'd like to manually add stable-open logic library, I don't need to have them bundled in development branch 

This is now just nitpicking on my side. You decide if you'll have stable+develop cores or just develop cores. Either case doesn't have any functional impact.

But in stable branch (`main`) I would like to se only one type of cores (stable)

src/base/olo_base.core
obruendl: Looks like a bug. IMO the paths are correct as per the [CAPI definition](https://fusesoc.readthedocs.io/en/stable/user/build_system/core_files.html#source-files):

image

I don't want to change paths to "src/base/vhdl" - because this is against the documented format of corefiles and will be broken if the bug in fusesoc is fixed.

The only quick fix I'd see is moving all corefiles to the top-level - but I want to avoid this. Having 10 (and potentially more in future) core-files in the top-level of the repo would be quite messy.

Could you report the bug and post a link here?  I would then follow that issue and ping Olof Kindgren.

obruendl:
The problem with keeping _main_ different (stable core-files only) from any other branch (development core-files only) is that these are branches of the same repo (containing the same files). How should this be realized? 

If I make a PR from _develop_ to _main_, code still is on _develop_ but goes to _main_ after being accepted. If I branch off from _main_ I am on some branch but I initially have the same codebase as in _main_ (including the core-files).

 I currently don't see how this should be implemented - in contrast to having both versions on all branches (which is easy to implement). If you see a straightforward way to implement the behavior you suggested, let me know.

src/base/olo_base.core
obruendl: Never mind, I reported it myself.

https://github.com/olofk/fusesoc/issues/717

Feel free to add a comment there.

obruendl: BTW: I am not sure how big chances is that the issue is reacted on quickly. FuseSoC currently has 141 open issues and since summer only two issues were closed.

rbrglez: > Looks like a bug. IMO the paths are correct as per the [CAPI definition](https://fusesoc.readthedocs.io/en/stable/user/build_system/core_files.html#source-files):

I don't think this is a bug. cores with `provider` key behave differently. I think that FuseSoC documentation should be updated to reflect that.

I have added explanation of the core behavior with `provider` key in this [repository](https://github.com/rbrglez/open-logic-example/tree/master) [README](https://github.com/rbrglez/open-logic-example/blob/master/README.md#creating-custom-remote-cores) file.

Here is an excerpt:

> ## Creating Custom Remote Cores
> 
> Check out `remote-cores/development/custom-open-logic/olo_fifo.core`.
> 
> `olo_fifo.core` is a custom core based on `olo_base.core`, but with a focus solely on FIFO files and their dependencies. It demonstrates an important behavior of core files managed by FuseSoC.
> 
> When a core file includes a `provider` key, it changes the way source files are resolved. Instead of searching for source files within the local directory of the core file itself, FuseSoC will download the entire repository specified in the provider key to the `~/.cache/fusesoc` directory.
> 
> From there, FuseSoC will reference and add source files starting from the top-level directory of the downloaded repository, not from the local directory where the core file resides.
> 
> This behavior is crucial to understand when working with core files that rely on external repositories for their source files, as it can affect how and where the source files are retrieved during the build process.
> 
> Previously, the behavior was somewhat unclear due to the existence of both `olo_base.core` and `olo_base_dev.core`.
> 
> - `olo_base.core` included a `provider` key, which caused FuseSoC to download the entire repository to `~/.cache/fusesoc`.
> - `olo_base_dev.core`, on the other hand, did not include the `provider` key, so it sourced files locally.
> 
> It's important to note that `olo_base.core` (the one with the `provider` key) has no direct connection to the `olo_base.core` file in the downloaded repository located in `~/.cache/fusesoc`. As a result, the core file **MUST** specify the full path from the root of the downloaded repository to the files it wants to include. 
> 
> **Why Did This Work Previously?**
> 
> You might wonder why this setup worked without errors even though the paths in olo_base.core were incorrect. When the repository was downloaded, FuseSoC initially attempted to locate the source files within the downloaded repository. When it couldn't find them there, it looked in the current directory.
> 
> However, this fallback mechanism didn't retrieve the correct files; instead, it found the development files. This is why the process appeared to work, but it was actually pulling in the wrong files — the development versions — instead of the intended production sources.


Only buggy behavior I see that cores with `provider` key download repository in `~/.cache/fusesoc` and when they don't find sources in downloaded repository they search from the current directory of the core.

rbrglez: > The only quick fix I'd see is moving all corefiles to the top-level - but I want to avoid this. Having 10 (and potentially more in future) core-files in the top-level of the repo would be quite messy.

I see a better fix: Don't include cores with `provider` key. Core files without `provider` key behave as expected. We have problems only with the cores that have `provider` key

rbrglez:
Ok, in that case I suggest that you drop cores with `provider` key from open-logic repository. What you could do is create documentation, how open-logic repository can be included in user projects with fusesoc. Something similar I made in this [repository](https://github.com/rbrglez/open-logic-example/tree/master), where I included open-logic with "remote-cores" and "local-cores"

rbrglez:
> I currently don't see how this should be implemented - in contrast to having both versions on all branches (which is easy to implement). If you see a straightforward way to implement the behavior you suggested, let me know.

**Example of a release strategy with only "local-cores" (no cores with `provider` key)**

1. The `develop` branch contains the core `open-logic:open-logic-dev:base:3.1.0`. Once the `develop` branch is reviewed and ready for release, create a pull request (PR) into a new branch, `pre-release`. NOTE that the bumping of version number happens in the `develop` branch, when you would know what kind of release you are preparing (patch, minor or major release)

2. In the `pre-release` branch, the only change you make is to modify the core filenames by removing the `-dev` suffix, resulting in `open-logic:open-logic:base:3.1.0`.

3. Afterward, create a PR into the `main` branch to finalize the release.

Key Points:

- The `develop` branch will always contain the `-dev` suffix in the Library of VLNV (e.g., `open-logic:open-logic-dev:base:3.1.0`.
- The `main` branch will always contain the official, production-ready VLNV (e.g., `open-logic:open-logic:base:3.1.0`).

obruendl:
Got it. The behavior for files with provider keys makes sense somehow - one could want to write a .core file for a 3rd party repo that does not contain core files and in this situation I get why paths can't be relative to the .core file. But the documentation really is misleading here.

Are you somehow involved in FuseSoC? If so, could you make sure the documentation is updated?

I found a nice solution and just pushed it. Please try.

The new setup is as follows:
- src/[area] contains the -dev .core files only
- tools/fusesoc/stable contains the stable core files (with paths relative to the repo root)

On my end this now works perfectly fine, for both the -dev version and the stable one. The setup has a nice side-effect: If you import "open-logic/src" instead of "open-logic" in fusesoc, you only get the -dev cores and the stable ones do not show up - which is what you would prefer as far as I understood.

BTW: What are you working on exactly? I am always interested to hear in what context people use Open Logic.

tools/fusesoc/stable/3.0.2/olo_axi.core
rbrglez, line 1: ok, this implementation is better than before, because it gives options to the users what will they include (development cores or release cores).

But I would still prefer that this cores with `provider` key would be just mentioned in documentation under instructions how to include open-logic with fusesoc.

How will this look in released `main` branch?
Will `olo_axi_dev.core` still have description that this is development version? If yes then I can come up with non-intuitive scenario of including  open-logic to user project with fusesoc.

Let's say that user will include open-logic like this:
```bash
fusesoc library add open-logic https://github.com/open-logic/open-logic
```

This will immediately   add all cores (local-development and remote-stable)

```bash
$ fusesoc core list

Available cores:

Core                                             Cache status  Description
================================================================================
open-logic:open-logic-dev:axi:3.0.2             :      local : local files (release plus WIP); AXI related modules see https://github.com/open-logic/open-logic/blob/main/doc/EntityList.md#axi
open-logic:open-logic-dev:base:3.0.2            :      local : local files (release plus WIP); Basic Circuitry (e.g. FIFOs, CDCs, ...) see https://github.com/open-logic/open-logic/blob/main/doc/EntityList.md#base
open-logic:open-logic-dev:intf:3.0.2            :      local : local files (release plus WIP); Interfaces (e.g. I2C, synchronizer, SPI, ...) see https://github.com/open-logic/open-logic/blob/main/doc/EntityList.md#intf
open-logic:open-logic:axi:3.0.2                 :      empty : official release (stable); AXI related modules see https://github.com/open-logic/open-logic/blob/main/doc/EntityList.md#axi
open-logic:open-logic:base:3.0.2                :      empty : official release (stable); Basic Circuitry (e.g. FIFOs, CDCs, ...) see https://github.com/open-logic/open-logic/blob/main/doc/EntityList.md#base
open-logic:open-logic:intf:3.0.2                :      empty : official release (stable); Interfaces (e.g. I2C, synchronizer, SPI, ...) see https://github.com/open-logic/open-logic/blob/main/doc/EntityList.md#intf
open-logic:tutorials-dev:quartus_tutorial:3.0.2 :      local : local files (release plus WIP); quartus tutorial for open-logic, targetting DE0-CV board
open-logic:tutorials-dev:vivado_tutorial:3.0.2  :      local : local files (release plus WIP); vivado tutorial for open-logic, targetting Zybo Z7-10 board
open-logic:tutorials:quartus_tutorial:3.0.2     :      empty : official release (stable); quartus tutorial for open-logic, targetting DE0-CV board
open-logic:tutorials:vivado_tutorial:3.0.2      :      empty : official release (stable); vivado tutorial for open-logic, targetting Zybo Z7-10 board

```

If user isn't privy to our discussion and reasoning he might think that `open-logic-dev` aren't stable cores, so he will use released stable cores which are remote.

While he is using open-logic files he will maybe change source files a little, but when he tries to use modified source code fusesoc won't take his changed source files, instead it will download fresh open-logic repository to `~/.cache/fusesoc` and user will lose time wondering why fusesoc won't use his modified source files

rbrglez:
> Got it. The behavior for files with provider keys makes sense somehow - one could want to write a .core file for a 3rd party repo that does not contain core files and in this situation I get why paths can't be relative to the .core file. But the documentation really is misleading here.

Yeah, I agree that documentation about `provider` cores needs to be improved

> Are you somehow involved in FuseSoC? If so, could you make sure the documentation is updated?

No, I only recently found out about this project and I am trying to use it at work

> I found a nice solution and just pushed it. Please try.
>
> On my end this now works perfectly fine, for both the -dev version and the stable one. The setup has a nice side-effect: If you import "open-logic/src" instead of "open-logic" in fusesoc, you only get the -dev cores and the stable ones do not show up - which is what you would prefer as far as I understood.

I've tried it. It is indeed much better than before, but I would still prefer If you didn't include cores with `provider` key. Check out my reasoning in my latest review comment.

> BTW: What are you working on exactly? I am always interested to hear in what context people use Open Logic.

Nothing serious yet. I've started advocating to my coworkers for the adoption of Open Logic library over our current proprietary company library, which seems to be unmaintained,  lacks automated tests, proper documentation, and any consistent coding standards.

I got a comment from one of my coworkers that he doesn't like the coding style you use, because you use variables in synthesizable code. His concern was that some tools might have problems with synthesizing this code. In previous company I have also used this Two process coding style with variables without problems, but I have used only vivado for synthesis.

tools/fusesoc/stable/3.0.2/olo_axi.core
obruendl: I see your scenario. 

But first, I think this is the smaller problem than if fusesoc resolves dependencies and interprets something as open-logic:xxx:3.0.2 which actually is not 3.0.2 but a local version with whatever changes or a "develop" branch state that contains non-backward compatible changes and will become 4.0.0 upon release.

And second, I am currently working on updating the fusesoc documentation in the [How To](https://github.com/open-logic/open-logic/blob/feature/fusesoc-dev/doc/HowTo.md#use-open-logic-through-fusesoc). It's not fully done yet - but the point that -dev means local files with changes and stable means downloaded from GitHub already is there. 

I clarified the wording in the description, so it explicitly says "downloaded from GitHub". 

With those two measures in place, I would assume no confusion for users that consulted the doc and read the core description. For those that don't read I guess confusion is unavoidable.

Definetly both options have their drawbacks but I rate "knowing that stable release 3.0.2 really is this release and downloaded from GitHub without changes" over "confusion of users who didn't read documentation". Probably our preferences differ here but I won't change mine due to experiences I had with library maintenance in the past.

BTW: Let your colleague who is concerned about variables know, that he can send me his exact concerns as linked-in message. I will happily follow up. I work since 15 years in the FPGA business targeting various vendors and I never ever had troubles with variables. Somehow people think that they are prone to latches - but you can accidentally build latches equally easily with signals. However, if your colleague has specific issues in mind that are known to exist in todays tools, I would of course be interested to know.

I am happy to hear that you are pushing the introduction of Open Logic. Let anybody with concerns know that they can contact me.

obruendl: I am done with the documentation as well. Please review it as well and let me know if it is clear of if I shall emphasize or clarify anything.

doc/HowTo.md
```diff
@@ -146,27 +146,72 @@ fuse-soc. To use Open Logic through fusesoc, just add open logic as a library:
...
 ```
 
-**Note:** Like for all other tool integrations, Open Logic entities are compiled into the library _olo_ and can be
-instantiated using e.g. `i_fifo : entity olo.olo_base_fifo_sync`.
+Packages are provided in two libraries:
+
+- Use **open-logic:open-logic** for **downloading the stable release from GitHub**
+- Use **open-logic:open-logic-dev** for using the **local files** (the release number mentioned plus all work in
+  progress which is not released)
+  - This option also allows using content of a feature branch that is checked out.
```
rbrglez, line 192: This is well-written and alleviates my concerns that the user will be confused about how to use Open-Logic with FuseSoC :)

rbrglez (approved):
this now looks good

rbrglez:
Could you maybe create another repository under open-logic account, where you would show different examples how to include open-logic with fusesoc to a dummy project?

Examples would include:
- using remote .core files without source code
- including open-logic as a git submodule and using stable or development cores

Essentially you could replicate this repository:
https://github.com/rbrglez/open-logic-example

obruendl:
I refined documentation, describing how to achieve different things by using Open Logic through fusesoc. 

I prefer not having to maintain another repo - especially because this is mostly FusesSoC usage know-how, which IMO shall be documented in FuseSoC and not in the context of Open Logic.

obruendl (commented):
reviewed

PR #125, opened 2025-03-11
Prevent TMR insertion in CDC blocks

svancau:


obruendl:
As discussed on Linked-In, this change is not added to the mainline because it for consistency TMR synthesis would have to be covered for all vendors, which is too much effort given the fact that it is not a very common need of users.

You are free to keep your fork but - at least at this point in time - I won't invest time to cover TMR synthesis in the mainline in the general case.

PR #132, opened 2025-03-23
FEATURE: Initial commit of PWM implementation.

rbrglez:


src/intf/vhdl/olo_intf_pwm.vhd
```diff
@@ -0,0 +1,203 @@
...
+    generic(
+        MaxPeriod_g : positive
+    );
+    port(
+        Clk : in std_logic;
+        Rst : in std_logic;
+
+        In_En     : in std_logic                                          := '1';
+        In_Period : in std_logic_vector(log2Ceil(MaxPeriod_g)-1 downto 0) := toUslv(MaxPeriod_g, log2Ceil(MaxPeriod_g));
+        In_Duty   : in std_logic_vector(log2Ceil(MaxPeriod_g)-1 downto 0);
```
obruendl: If "MaxPeriod_g=8" means the period can be up to 8 clock cycles, I think it would be most logical to have "In_Period" in the range between 0 and 8. Hence In_Period should be a:

std_logic_vector(log2ceil(MaxPeriod_g+1)-1 downto 0) 

Otherwise the value 8 is not possible.

The same range should apply to "In_Duty".

src/intf/vhdl/olo_intf_pwm.vhd
```diff
@@ -0,0 +1,203 @@
...
+            --------------------------------------------------------------------
+            when Idle_s =>
+                v.Pwm := '0';
+
+                v.PeriodCnt := (others => '0');
+
+                if (In_En = '1') then
+                    -- Decide which state is next
+                    -- (Edge case: Out_Pwm always low)
+                    if (unsigned(In_Duty) * unsigned(In_Period) < MaxPeriod_g) then
```
obruendl: Why the multiplication? Multiplications are expensive.

In_Duty to my understanding defines for how many clock-cycles the PWM is on. Correct? In this case I could not expect any multiplications.

Probably naming wise "In_Duty" could be changed to "In_OnTime" (to be clear that this is in clock cycles and not in %)?

src/intf/vhdl/olo_intf_pwm.vhd
```diff
@@ -0,0 +1,203 @@
...
+        variable v : TwoProcess_r;
+    begin
+        -- *** hold variables stable ***
+        v := r;
+
+        -- *** Default Values ***
+        v.PeriodStart := '0';
+
+        -- *** State Machine ***
+        case (r.State) is
```
obruendl: Is there are need for an FSM?

Wouldn't it be much less code if r.PeriodCnt is just running from 0 to In_Period-1 and PWM is on whenever r.PeriodCnt < In_Duty?

I agree with stopping the counter when In_En = '0'. However, it would be simple to add the handling of "In_En='0'" as simple if-statement at the end of the process. The "if" already is there, but the idea of just overriding all registers there is not yet implemented this way.

src/intf/vhdl/olo_intf_pwm.vhd
```diff
@@ -0,0 +1,203 @@
...
+                v := TwoProcessInit_c;
+        end case;
+
+        if (In_En = '0') then
+            v.Pwm := '0';
+            --
+            v.State := Idle_s;
+        end if;
+
+        if (Rst = '1') then
```
obruendl: Functionally I have nothing against  putting the reset into the combinatorial process.

However, for consistency reasons: Could you put it into the sequential process (like it is the case for most other existing components)?

obruendl (commented):
I added some comments.

src/intf/vhdl/olo_intf_pwm.vhd
rbrglez: Done

src/intf/vhdl/olo_intf_pwm.vhd
rbrglez: I have used multiplication, because In_Duty unit was percents, not clock cycles.

Now I have Renamed In_Duty to In_OnTime, which simplified the design significantly.

src/intf/vhdl/olo_intf_pwm.vhd
rbrglez: Removed State machine

src/intf/vhdl/olo_intf_pwm.vhd
```diff
@@ -0,0 +1,117 @@
...
+        r_next <= v;
+
+    end process;
+
+    -- Outputs
+    Out_PeriodStart <= '1' when (r.PeriodCnt = 0) and (In_En = '1') else '0';
+    Out_PeriodCnt   <= std_logic_vector(r.PeriodCnt);
+    Out_Pwm         <=
+        '1' when (r.PeriodCnt < r.OnTime) and (r_next.OnTime > 0) and (In_En = '1') else
+        '0';
```
rbrglez: I have tested this entity in my local non-vunit testbench and I am quite happy with the results. However I am not particularly  happy with "Outputs" section of the entity.

Is it OK that I am using `In_En = '1'` and `r_next.OnTime > 0` in `when-else`

src/intf/vhdl/olo_intf_pwm.vhd
```diff
@@ -0,0 +1,117 @@
...
+        v.PeriodStart := '0';
+
+        if (v.PeriodCnt = to_unsigned(0, r.PeriodCnt'length)) then
+            -- Update values for Period and OnTime
+            v.Period := unsigned(In_Period);
+            v.OnTime := unsigned(In_OnTime);
+        end if;
+
+        if (In_En = '1') then
+            if (v.Period = to_unsigned(0, r.Period'length)) then
```
rbrglez: Q1)
Is it OK that i have variable (`v.Period`) in if statement?

Q2)
Is it good practice that I am converting integer to unsigned or it would be fine if i just compared unsigned to integer?
Alternative:
```vhdl
if (v.Period = 0) then
```

src/intf/vhdl/olo_intf_pwm.vhd
obruendl: Ideally outputs should be registered. 

Adding a register stage to all outputs seems trivial at first - but thinking about it in more detail, it somehow makes _Out_PeriodStart_ useless (I would assume this signals is mostly used to know when inputs are latched).

I'd see the following as sane trade-off:
- Register _Out_Pwm_ 
- Register _Out_PeriodCnt_ (this I'd assume to be used e.g. to synchronize ADC measurements)
- Remove _Out_PeriodStart_ and instead add _In_Ready_ (pulsed high in the clock cycles the inputs are latched)

What's your opinion on this? I think there is no "ideal" solution here - it is about finding a trade-off that works best for most cases - so I am very open for other opinions.

src/intf/vhdl/olo_intf_pwm.vhd
obruendl: **For Q1**

Generally, in reading v.Xxx values should be avoided because it can lead to long combinatorial paths. I do NOT prohibit it - but when it is done it should be carefully discussed (as you do here -> perfect).

The logic here is simple enough to accept it when the input was known to be a register. However, in this case it creates a longer combinatorial path from an input (In_Period) to a regisetr (r.PeriodCnt) - and the input we do not have under control, the user may apply something that already has a long combinatorial path.

Still I'd accept it here if there is no simple alternative - I don't expect this to ever become really critical.

Regarding _simple alternative_: What's the reason for explicitly checking it? By removing the check, _In_Period=0_ would always end up in the else clause, which at least on a first gance would seem fine to me.

**For Q2**

`v.Period = 0` is clearly to be preferred. It leads to exactly the same synthesis results, but is more readable, more consise and more safe (comparing vectors has corner cases like always returning false if lenghts do not match).

obruendl (commented):
NExt round doune

obruendl:
@rbrglez 

Any update on this one? Is it still active?

obruendl:
@rbrglez 

I feel like with the production code we might not be far from where it is but a few touches still are required. 

Would it be okay to you if I just grab the pwm source file and add the last things I think are needed?

Regarding testbench: How familiar are you with VUnit? I think for this one as well as for other functionality we should have a VUnit Verification Component that allwos checking the pattern on a binary signal (like "high for between x and y sec, then low for between u and v sec, ...". Would you feel comfortable to write such a verification component or would you prefer I do it?

PR #134, opened 2025-03-27
FEATURE: Allow BitsPerSymbol_g > LFSR Width for olo_base_prbs

obruendl:


obruendl:
@hh44bbbbyy 

I applied my changes. I also decided that I want to avoid that the upper bits in the first produced symbol are all zero - instead in the BitsPerSymbol > LFSR-Width case I now shift the seed before producing the first symbol, so all bits in the first symbol are belonging to the LFSR sequence.

Please review and approve if you agree (or leave comments otherwise).

PR #156, opened 2025-05-30
new module development: incremental encoder interface

MiloradPetrovic1:
New entity that is an interface to an incremental encoder.

### DIFF
--- a/test/intf/olo_intf_inc_encoder/olo_intf_inc_encoder_tb.vhd
+++ b/test/intf/olo_intf_inc_encoder/olo_intf_inc_encoder_tb.vhd
@@ -142,7 +142,7 @@ begin
 
         end procedure;
 
-        variable ExpectedAngleValue : unsigned(63 downto 0);
+        variable TmpU64 : unsigned(63 downto 0);
 
     begin
         test_runner_setup(runner, runner_cfg);
@@ -185,8 +185,8 @@ begin
                     ClockCyclesPerStep => 10);
 
                 check_equal(DUT_Position_Value, toUslv(10_000, DUT_PositionWidth_g), "DUT 'Position_Value' is not the expected value after the positive direction rotation.");
-                ExpectedAngleValue := UnsignedMod(to_unsigned(10_000, 64), EmulatedEncoder_Resolution_g);
-                check_equal(DUT_Angle_Value, ExpectedAngleValue(DUT_Angle_Value'range), "DUT 'Angle_Value' is not the expected value after the positive direction rotation.");
+                TmpU64 := UnsignedMod(to_unsigned(10_000, 64), EmulatedEncoder_Resolution_g);
+                check_equal(DUT_Angle_Value, TmpU64(DUT_Angle_Value'range), "DUT 'Angle_Value' is not the expected value after the positive direction rotation.");
 
                 -- Rotate encoder down for some amount.
                 RotateEncoder(
@@ -195,8 +195,8 @@ begin
                     ClockCyclesPerStep => 10);
 
                 check_equal(DUT_Position_Value, toUslv(8_000, DUT_PositionWidth_g), "DUT 'Position_Value' is not the expected value after the negative direction rotation.");
-                ExpectedAngleValue := UnsignedMod(to_unsigned(8_000, 64), EmulatedEncoder_Resolution_g);
-                check_equal(DUT_Angle_Value, ExpectedAngleValue(DUT_Angle_Value'range), "DUT 'Angle_Value' is not the expected value after the negative direction rotation.");
+                TmpU64 := UnsignedMod(to_unsigned(8_000, 64), EmulatedEncoder_Resolution_g);
+                check_equal(DUT_Angle_Value, TmpU64(DUT_Angle_Value'range), "DUT 'Angle_Value' is not the expected value after the negative direction rotation.");
 
             end if;
 
@@ -287,6 +287,60 @@ begin
 
                 -- DUT 'Angle_Value' should be zero.
                 check_equal(DUT_Angle_Value, to_unsigned(0, DUT_Angle_Value'length), "DUT 'Angle_Value' was not zero as expected.");
+
+            end if;
+
+            if run("ClearAngle") then
+
+                -- Reset DUT, and then release reset.
+                WaitForClockCycles(Clk, 1);
+                DUT_Rst <= '0';
+
+                -- Rotate upwards for 90 degrees.
+                TmpU64 := "00" & EmulatedEncoder_Resolution_g(EmulatedEncoder_Resolution_g'high downto 2);
+                RotateEncoder(
+                    StepDirection      => STEP_UP,
+                    NumberOfSteps      => TmpU64,
+                    ClockCyclesPerStep => 10);
+
+                -- Validate 'DUT' state.
+                check_equal(DUT_Angle_Value, TmpU64(DUT_Angle_Value'range), "DUT 'Angle_Value' does not correspond to 90 degrees.");
+
+                -- Clear the DUT angle counter.
+                DUT_Angle_Clear <= '1';
+                WaitForClockCycles(Clk, 1);
+                DUT_Angle_Clear <= '0';
+                WaitForClockCycles(Clk, 1);
+
+                -- Validate 'DUT' state.
+                check_equal(DUT_Angle_Value, to_unsigned(0, DUT_Angle_Value'length), "DUT 'Angle_Value' is not zero after clear.");
+
+            end if;
+
+            if run("ClearPosition") then
+
+                -- Reset DUT, and then release reset.
+                WaitForClockCycles(Clk, 1);
+                DUT_Rst <= '0';
+
+                -- Rotate upwards for half the position range.
+                TmpU64 := (DUT_PositionWidth_g - 2 => '1', others => '0');
+                RotateEncoder(
+                    StepDirection      => STEP_UP,
+                    NumberOfSteps      => TmpU64,
+                    ClockCyclesPerStep => 10);
+
+                -- Validate 'DUT' state.
+                check_equal(DUT_Position_Value, TmpU64(DUT_Position_Value'range), "DUT 'Position_Value' does correspond to 90 degrees.");
+
+                -- Clear the DUT position counter.
+                DUT_Position_Clear <= '1';
+                WaitForClockCycles(Clk, 1);
+                DUT_Position_Clear <= '0';
+                WaitForClockCycles(Clk, 1);
+
+                -- Validate 'DUT' state.
+                check_equal(DUT_Position_Value, to_unsigned(0, DUT_Position_Value'length), "DUT 'Position_Value' is not zero after clear.");
 
             end if;
 


src/base/vhdl/olo_base_pkg_array.vhd
```diff
@@ -26,6 +26,8 @@ library ieee;
 ---------------------------------------------------------------------------------------------------
 package olo_base_pkg_array is
 
+    type SlvArr_t is array (natural range <>) of std_logic_vector;
```
obruendl, line 29: I explicitly removed arrays of unconstrained vectors because Quartus Prime Standard does not allow this. Even worse - it will fail to compile the package if the type is nowhere used.

Hence - please remove this one unless Quartus Prime Standard behavior is confirmed to have changed.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,42 @@
...
+
+entity olo_intf_inc_encoder is
+    generic(
+        PulsesPerPhasePerTurn : positive;
+        UpCountingDirection : string := "B trails A");
+    port(
+        Clk             : in  std_logic;
+        Rst             : in  std_logic;
+        Phase_A         : in  std_logic;
+        Phase_B         : in  std_logic;
```
obruendl: Can we add Z (index)?

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,42 @@
...
+    use ieee.std_logic_1164.all;
+    use ieee.numeric_std.all;
+
+library work;
+    use work.olo_base_pkg_math.all;
+
+entity olo_intf_inc_encoder is
+    generic(
+        PulsesPerPhasePerTurn : positive;
+        UpCountingDirection : string := "B trails A");
```
obruendl: I'd suggest to keep the counting direction constant - but cleanly document it. If somebody does not agree on the counting direction, he can simply flip the phases.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,42 @@
...
+library ieee;
+    use ieee.std_logic_1164.all;
+    use ieee.numeric_std.all;
+
+library work;
+    use work.olo_base_pkg_math.all;
+
+entity olo_intf_inc_encoder is
+    generic(
+        PulsesPerPhasePerTurn : positive;
```
obruendl: Do we need to know the resolution of the encoder if we only count the position?

As far as I understand, we would only need it if we want to count the angle (which wraps after one rotation).

I actually would find this a good feature becasue the angle is often required for commutation. Could we add the _Angle_ counting?

Regarding the encoder resolution: In some cases the same logic will be used for different motors with different encoders. Having the encoder resolution as generic is a bit limiting here. However, adding it as a port is cumbersome if it is fixed and one just wants to assign an integer.

Could we do as follows:
- Add a generic `DefaultEncoderLines_g : positive`
- Add a port `EncoderLines : std_logic_vector(PositionWidth_g-1 downto 0) := toUslv(DefaultEncoderLines_g, PositionWidth_g)`

This way the user could decice if he wants to have a fixed encoder resolution or if he wants to change it at runtime.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,42 @@
...
+        UpCountingDirection : string := "B trails A");
+    port(
+        Clk             : in  std_logic;
+        Rst             : in  std_logic;
+        Phase_A         : in  std_logic;
+        Phase_B         : in  std_logic;
+        Strobe          : in  std_logic := '0';
```
obruendl: What is "strobe"?

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: Can we add _PositionWidth_g_ ? 

One could have an encoder with only 512 lines but still wanting to count the position as 32-bit value.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,42 @@
...
+        Position_tdata  : out std_logic_vector(log2ceil(PulsesPerPhasePerTurn * 4) - 1 downto 0);
+        Position_tvalid : out std_logic;
+        Position_tready : in  std_logic := '0';
```
obruendl: I'd suggest to not make _Position_ an AXI4-Stream but keep it a simple parallel output. Reason: The position is always valid. And I also would find a tready very counterintuitive - the encoder continues turning, it's unclear what backpressure would mean in this context.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,42 @@
...
+        Clear           : in  std_logic := '0';
```
obruendl: What's the difference between _Rst_ and _Clear_?

Is there any reason why the user should not use the existing _Rst_ input to clear the counter?

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: Note - Saw it was just further below. Hence the unhandshaked _Position_ exist - I'd suggest to remove the AXI4-Stream handshaked version - unless there is a good reason for it I did overlook (in this case, let me know).

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,42 @@
...
+        Pulse           : out std_logic;
+        Direction       : out std_logic;
```
obruendl: It probably does not matter much - but could we go with "Pulse_Up" and "Pulse_Down"? In contrast to "Direction" their meaning is implicitly clear (for Direction, it's not implicitly clear if '1' means _up_ or _down_).

src/base/vhdl/olo_base_pkg_array.vhd
MiloradPetrovic1: Please give me a day or two to install Quartus Prime and play around with this issue. I do not have the license for Quartus Prime Standard but I assume that all will be the same, regarding this issue, as for Quartus Prime Lite?

Without this type (array of SLV), a lot of flexibility will be lost, not for this encoder entity, but for the olo project in general.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: We can of course add the Z phase, however, the Z phase is very application specific. So what do we want to do with the Z phase? Reset the internal counter? Check if the value of the counter is the same as when the last Z pulse occurred (throw an error otherwise)?

I would leave the Z phase logic to the end user. But if you have any requirements, please share.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Yes, what you say makes sense. I will do it so.

Also, to this end, I think it is good to rename the ports Phase_A and Phase_B to Phase_1 and Phase_2. This might reduce confusion: Up counting will result when the signal that is hooked up to Phase_1 (be it the 'physical' phase A or B of the encoder) is leading the signal that is hooked up to Phase_2. Do you agree?

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: My idea was when there is a pulse on the *strobe* port, the current value of the encoder counter gets put on the output AXI-Stream bus.

This makes sense because for example in motor control applications, you want to sample the position of the rotor in sync with the running of the speed/position control loop or with the PWM modulator time-base. I have done some work with FPGA applications for power electronics and motor control, so I would say this syncing is important.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: *Clear* would just clear the internal encoder position counter, but *Rst* would reset the entire module.

So a *Clear* can for example be used when zeroing the encoder.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Yes, you are right, I will do it like that. It had not occurred to me but *PulseUp* and *PulseDown* would be a more elegant solution.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: I would keep the AXI-Stream interface for reasons explained regarding the *Strobe* port. So we want to latch a measurement at a specific time or event. I think it would be elegant to put that latched measurement on an AXI-Stream bus, so it can be used in other parts ot the design. Your thoughts?

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: My idea, as stated in the comments above, was that the width of the position register is based on the generic *PulsesPerPhasePerTurn*. The relation holds: *PositionWidth_g = ceil(log2(4 x PulsesPerPhasePerTurn*. But the value of *PulsesPerPhasePerTurn* is usually what is in the datasheet of the encoder.

I dont really understand why someone would use a bigger encoder counter register than needed? Padding zeros can always be added external to this entity to make the data conform to a certain format (like a 32-bit value).

MiloradPetrovic1: My idea was to calculate the minimal width of the encoder counter register needed based on the encoder resolution. So if one phase has a total of **N** pulses per full turn, both phases will generate a total of **4 x N** pulses per full turn. Hence a register of width **ceil(log2(4 x N))** is needed.

I am not quite sure that I understand the rest of your comment?
- The encoder resolution is fixed for one encoder that is hooked up on a motor. The need fo dynamically configuring the encoder resolution is **highly** unlikely.
- No quite sure what you mean by *angle counting*? Do you want a way to track the number of full rotations? That would definitely be useful but I would leave that as a functionality implemented external to this entity.

src/base/vhdl/olo_base_pkg_array.vhd
obruendl: Quartus prime lite is ok. Thats the device limited version of quartus prime.

I agree that losing this type is a pity. But I do not want to break compatibility with the tools of the second largest vendor. Thats the price I pay for being fully vendor independent.

Note: I believe quartus prime standard will be replaced by quartus prime pro. In some 5 or the like support for quartus prime standard can be droped. So at least the restrictions do not aply forever.

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: Z is the index. A pulse on z shall reset the angle (not the position)

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: I'd still go with A and B. Their meaning still is the same and the ability to flip them should be widely known.

src/base/vhdl/olo_base_pkg_array.vhd
obruendl: BTW: Please check if quartus Prime accepts the statement when the code is compiled as VHDL 2008.

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: Often quadrature encoders are used on servo motors. In this case they measure two things.

* The Angle. This counter is reset after every rotation and hence we need exactly 4 x EncoderLines as range.
* The Position. A servo-drive may move 2 cm per motor rotation but cover 2 m in distance (requiring 100 rotations) from one end to the other.

For the angle the width you calculated is fine. For the position, the user should give the width required - often users might chose the width of some register interface but they could also adjust it according to special needs.

Regarding "dynamic changing of encoder resolution". Yes, of course the encoder attached to a motor is given. But it is not uncommon that the same FPGA bitstream can be used for different motors - think of an OTS servo controller where the user can attach any motor with any encoder and then configure the encoder resolution. In such a scenario the encoder resolution is not known at compile-time of the FPGA bitstream.

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: IMO we should keep separate things apart. What you describe here is sampling of it (which would be the same concept for any other values).

I'd suggest to make the encoder interface just convert encoder to position/angle - and implement the "AXI-Stream-Sampler" separately if really needed. That way the sampler could be used on any data - not only on encoders.

I am not sure if a sampler makes sense though - usually control loops just take the current position upon a strobe. This is a one-liner in HDL and I somewhat question if anybody would want to instantiate an entity for this.

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: As mentioned above: I'd not put a sampler inside the encoder interface. We should just convert A/B/(Z) to Angle/Position. Sampling can be done independently.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Ok, now I understand what you mean by *position*, *angle* and the need to dynamically configure the resolution.

I will make it as you said.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Ok, I understand, we want the sampling separately. Will do.

MiloradPetrovic1:
I have made the requested changes. Please let me know if the entity declaration is now what you expected.

obruendl:
Looks OK now.

Implementation, testbench and documentation are next then. Let me know when there is anything ready for review.

Before you start into implementation: Do you have a good coding setup in place? Especially having the linter and running simulation for the current file on a hotkey in VSCode helps a slot, also the navigation options of VHDL-LS I appreciate. I described how to set-up those things in the [HowTo](https://github.com/open-logic/open-logic/blob/main/doc/HowTo.md). Let me know if you need any help in setting things up.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+
+architecture rtl of olo_intf_inc_encoder is
+    
+    -- The values of the encoder phase signals from the prior clock cycle.
+    signal Phase_A_Prior : std_logic;
+    signal Phase_B_Prior : std_logic;
+    signal Phase_Z_Prior : std_logic;
+
+    -- Encoder position value as a numeric_std.unsigned.
+    signal Position_U : unsigned(Position'range);
```
obruendl: I'd suggest to remove the type suffix _U. It's not exactly prohibited but it might confuse others because it is not in the guidelines,

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl, line 1: I think a olo_intf_sync shouldbe added on the encoder inputs (that would also be consistent with the existing olo_intf_... entities).

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+entity olo_intf_inc_encoder is
+    generic(
+        DefaultAngleResolution_g : positive;
+        PositionWidth_g   : positive := 32);
+    port(
+        Clk             : in  std_logic;
+        Rst             : in  std_logic;
+        Phase_A         : in  std_logic;
+        Phase_B         : in  std_logic;
+        Phase_Z         : in  std_logic := '0';
```
obruendl: Could we call them Enc_...?

I think it would emphasize what is connected here.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+        Position        : out std_logic_vector(PositionWidth_g - 1 downto 0);
+        Clear_Position  : in  std_logic := '0';
```
obruendl: Pos_Position and Pos_Clear? To emphasize those are related.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+        Angle           : out std_logic_vector(PositionWidth_g - 1 downto 0);
+        Clear_Angle     : in  std_logic := '0';
```
obruendl: Ang_Angle and Ang_Clear?

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+    Angle <= std_logic_vector(Angle_U);
+    
+    proc_main : process is
+        variable IncrementEventOccurred : boolean;
+        variable DecrementEventOccurred : boolean;
+        variable IndexEventOccurred     : boolean;
+    begin
+
+        -- Wait for an active clock edge.
+        wait until rising_edge(Clk);
```
obruendl, line 97: Is this synthesizable with all tools? 

I newer saw it that way in synthesizable code, so I am not sure. Any readon for not using the standard "if rising_edge()" clause?

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+
+        elsif IncrementEventOccurred then
+            Position_U <= Position_U + 1;
+
+        elsif DecrementEventOccurred then
+            Position_U <= Position_U - 1;
+
+        end if;
+
+        -- Update angle counter.
```
obruendl: I'd suggest to merge the if structures for angle and position. Its the same counting logic, so it can be merged.

The z logic that only exists for the angle can be added after the merged clunting logic as override.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+
+        elsif IndexEventOccurred then
+            Angle_U <= (others => '0');
+
+        elsif IncrementEventOccurred and DecrementEventOccurred then
+            Angle_U <= Angle_U;
+
+        elsif IncrementEventOccurred then
+            Angle_U <=
+                (others => '0') when Angle_U = unsigned(AngleResolution)
```
obruendl: That should be >= to cover the edge case where the angle resolution is reduced.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+        Clk             : in  std_logic;
+        Rst             : in  std_logic;
+        Phase_A         : in  std_logic;
+        Phase_B         : in  std_logic;
+        Phase_Z         : in  std_logic := '0';
+        Position        : out std_logic_vector(PositionWidth_g - 1 downto 0);
+        Clear_Position  : in  std_logic := '0';
+        Angle           : out std_logic_vector(PositionWidth_g - 1 downto 0);
+        Clear_Angle     : in  std_logic := '0';
+        AngleResolution : in  std_logic_vector(PositionWidth_g - 1 downto 0) := toUslv(DefaultAngleResolution_g, PositionWidth_g);
```
obruendl: It's a bitcawkward that the width is driven by the default value. Would it make sens3 to name it "MaxAngleResolution_g" and document that it also is the default?

I am veryopen for better ideas here...

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+
+        -- Update angle counter.
+        if Clear_Angle = '1' then
+            Angle_U <= (others => '0');
+
+        elsif IndexEventOccurred then
+            Angle_U <= (others => '0');
+
+        elsif IncrementEventOccurred and DecrementEventOccurred then
+            Angle_U <= Angle_U;
```
obruendl: I'd find it more readable to remove this and add "IncrementOcurred and not DecrementOcurred" to the counting detection.

Reason: At a first glance this clause looks needless because its content does not do anything - but it does prevent the counting branches from being reached. This functionally is correct butnot obvious for the reader.

src/intf/vhdl/olo_intf_inc_encoder.vhd
```diff
@@ -0,0 +1,149 @@
...
+
+begin
+
+    Position <= std_logic_vector(Position_U);
+    Angle <= std_logic_vector(Angle_U);
+    
+    proc_main : process is
+        variable IncrementEventOccurred : boolean;
+        variable DecrementEventOccurred : boolean;
+        variable IndexEventOccurred     : boolean;
```
obruendl: Add _v suffix - the linter should repot this as well.

test/intf/olo_intf_inc_encoder/olo_intf_inc_encoder_tb.vhd
obruendl, line 1: Didn't review it because it is not close to final

obruendl (changes requested):
Looks mostly ok. Most comments are style and not functional.

Have you already installed thevsg linter? I ask becauseI noticed the missing _v suffixes for variables, which it should report.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: My intention was that that _U just means the *unsigned* value of the *Position* port. How would you name such a signal?

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: We could do as you have said. But I think that such input pin manipulations should be done externally to the module. Because what if you want to invert the signals? What if you want to add some glitch removing logic? The same goes with syncing the input signal to the clock. So my assumtion was the this entity has on its inputs proper signals (clean and synchronized). Let me know what you think.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Yes, will do.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: How about **Position_Value** and **Position_Clear**?

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: How about **Angle_Value** and **Angle_Clear**?

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: This construct should be good with AMD tools, from my experience. As for other tools, I do not know for a fact but I can investigate.

The reasons why I prefer *wait until* over *if then* are the following:
- One less level of indentation.
- No need to add a process sensitivity list.
- Makes in definitive that the process is synchronous to the given clock.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Understood, will do.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Yes, will do.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: True, however you are assuming that the *AngleResolution* will change its value during operation **after** reset. The way I see it, the *AngleResolution* will be first set to a value (by some other FPGA logic) and only then will the entity be taken out of reset. So, during operation the *AngleResolution* value will remain constant. Am I missing something?

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: I will get back to you on this...

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Ok, will do.

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Surprisingly, the linter did **not** catch this specific issue, the missing **_v**?

In my terminal I executed:

`vsg -c ./lint/config/vsg_config.yml -f ./src/intf/vhdl/olo_intf_inc_encoder.vhd`

Am I maybe missing something? Or maybe you can try on your side?

src/intf/vhdl/olo_intf_inc_encoder.vhd
MiloradPetrovic1: Understood, will do.

obruendl:
Any update on this one?

MiloradPetrovic1:
I have made a more detailed testbench, and have made some changes to the entity. There is still a little more work I think is required, I will try to get one more commit in by tomorrow evening.

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: I do not use type suffixes - I would name it Position. 

If I have the same signal as Unsigned and std_logic_vector I probably would name it PositionSlv and PositionU or the like.

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: That's somewhat of a tradeoff - either you urge people to instantiate two entities or you are at risk of adding 2 FF stages which are not required (if the user adds an external olo_intf_sync + his inverter or whatever).

I'd still vote for adding the olo_intf_sync. For two reasons:
1. It is how all other intf elements do. Consistency more often outrules slight tehcnical details IMO.
2. 2 FF stages usually do not hurt - even if they should not be required

However, I technically agree that there are corner cases where the synchronizer is not required. For me the four FFs do not count enough to add a generic. If you can provide a good reason why adding a needless synchronizer hurts - let me know. In this case I would suggest we add the option to omit the synchronizer to all olo_intf entities (that way it'd be consistent again).

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: That's fine for me.

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: Thats fine for me.

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: Your assumption may be as wrong as mine. But if mine is wrong nothing bad happens ;-)

I tend to follow defensive coding. Normally users do any kind of unexpected things - so better be prepared for it.

Also: How much after the reset would the user have to set the angle? What happens if during that time the encoder has increments (or decrements... in that situation the AngleResolution really matters, even for a single decrement).

src/intf/vhdl/olo_intf_inc_encoder.vhd
obruendl: Let me try this out on my machine (I am currently not at home). Keep the comment open until I did.

PR #158, opened 2025-06-01
FEATURE: stream split/merge

MiloradPetrovic1:
The implementation of modules for AXI-Stream splitting and merging.

### DIFF
--- a/src/axi/vhdl/olo_base_stream_merge.vhd
+++ b/src/axi/vhdl/olo_base_stream_merge.vhd
@@ -18,23 +18,25 @@ library ieee;
 ---------------------------------------------------------------------------------------------------
 -- Entity
 ---------------------------------------------------------------------------------------------------
-entity olo_axi_stream_merger is
+entity olo_base_stream_merge is
     generic(
-        NumberOfInputStreams_g : positive := 2);
+        Inputs_g : positive := 2);
     port(
-        -- input streams
-        InputStreams_tvalid : in  std_logic_vector(NumberOfInputStreams_g downto 0);
-        InputStreams_tready : out std_logic;
-        -- output stream
-        OutputStream_tvalid : out std_logic;
-        OutputStream_tready : in  std_logic);
-end entity olo_axi_stream_merger;
+        -- Input streams handshake signals.
+        In_Valid : in  std_logic_vector(Inputs_g - 1 downto 0);
+        In_Ready : out std_logic;
+        -- Output stream handshake signals.
+        Out_Valid : out std_logic;
+        Out_Ready : in  std_logic);
+end entity;
 
-architecture rtl of olo_axi_stream_merger is
+architecture rtl of olo_base_stream_merge is
+    signal AllInputsValid : std_logic;
 begin
-    OutputStream_tvalid <= '1' when InputStreams_tvalid = (others => '1') else '0';
-    InputStreams_tready <= OutputStream_tvalid and OutputStream_tready;
-end architecture rtl;
+    AllInputsValid <= '1' when In_Valid = (others => '1') else '0';
+    Out_Valid <= AllInputsValid;
+    In_Ready <= AllInputsValid and Out_Ready;
+end architecture;
 
 
 


src/axi/vhdl/olo_axi_stream_merger.vhd
```diff
@@ -0,0 +1,40 @@
...
+---------------------------------------------------------------------------------------------------
+-- Libraries
+---------------------------------------------------------------------------------------------------
+library ieee;
+    use ieee.std_logic_1164.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity
+---------------------------------------------------------------------------------------------------
+entity olo_axi_stream_merger is
```
obruendl: I'd suggest to name this _olo_base_stream_merge_

The AXI area I reserved for AXI4 things. I asked myself often if AXI-Stream belongs in there - but because this would mean more or less all things of olo_base would go into olo_axi (they all have AXI-4 Stream handshaking), I decided that a simple ready/valid handshaking does not qualify for entities going into olo_axi.

src/axi/vhdl/olo_axi_stream_merger.vhd
```diff
@@ -0,0 +1,40 @@
...
+    generic(
+        NumberOfInputStreams_g : positive := 2);
```
obruendl: Can we make this more concise? e.g. _NoInputs_g_ or "_Inputs_g_ or _InStreams_g_ or the like?

src/axi/vhdl/olo_axi_stream_merger.vhd
```diff
@@ -0,0 +1,40 @@
...
+    port(
+        -- input streams
+        InputStreams_tvalid : in  std_logic_vector(NumberOfInputStreams_g downto 0);
+        InputStreams_tready : out std_logic;
+        -- output stream
+        OutputStream_tvalid : out std_logic;
+        OutputStream_tready : in  std_logic);
```
obruendl: Two things for the declarations here:

1. Please use suffixes "_Valid" and "_Ready". This is how it is everywhere else and I want it to be consistent and according to the [conventions](https://github.com/open-logic/open-logic/blob/main/doc/Conventions.md#axi4-stream-handshaking-signals). Could be that it would have been cleverer to call them TReady and TValid according to the AXI4-Stream spec - but changing it everywhere would be a huge breaking change and users seem to accept the Ready/Valid names well, so I think for the time being we should stay with it.

2. Try make names more concise. I think _In_xxx_ and _Out_xxx_ would do just fine.

src/axi/vhdl/olo_axi_stream_merger.vhd
```diff
@@ -0,0 +1,40 @@
...
+end entity olo_axi_stream_merger;
+
+architecture rtl of olo_axi_stream_merger is
+begin
+    OutputStream_tvalid <= '1' when InputStreams_tvalid = (others => '1') else '0';
+    InputStreams_tready <= OutputStream_tvalid and OutputStream_tready;
```
obruendl: You are reading an output _OutputStream_tvalid_ here. This won't work. 

It's easiest to add an internal version of the "AllInputsValid" signal to use locally.

src/axi/vhdl/olo_axi_stream_splitter.vhd
```diff
@@ -0,0 +1,37 @@
...
+---------------------------------------------------------------------------------------------------
+-- Libraries
+---------------------------------------------------------------------------------------------------
+library ieee;
+    use ieee.std_logic_1164.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity
+---------------------------------------------------------------------------------------------------
+entity olo_axi_stream_splitter is
```
obruendl, line 21: Same comment as for the merger. Should go into _olo_base_

src/axi/vhdl/olo_axi_stream_splitter.vhd
```diff
@@ -0,0 +1,37 @@
...
+    generic(
+        NumberOfOutputStreams_g : positive := 2);
```
obruendl, line 23: Same comment as for the merger: should be more concise

src/axi/vhdl/olo_axi_stream_splitter.vhd
```diff
@@ -0,0 +1,37 @@
...
+    port(
+        -- input stream
+        InputStream_tvalid : in  std_logic;
+        InputStream_tready : out std_logic;
+        -- output streams
+        OutputStreams_tvalid : out std_logic;
+        OutputStreams_tready : in  std_logic_vector(NumberOfOutputStreams_g - 1 downto 0));
```
obruendl, line 30: Same comment as for the merger: More concise and "Valid/Ready" instead of "TValid/TReady".

src/axi/vhdl/olo_axi_stream_splitter.vhd
```diff
@@ -0,0 +1,37 @@
...
+end entity olo_axi_stream_splitter;
+
+architecture rtl of olo_axi_stream_splitter is
+begin
+    InputStream_tready <= '1' when OutputStreams_tready = (others => '1') else '0';
+    OutputStreams_tvalid <= InputStream_tvalid and InputStream_tready;
```
obruendl, line 36: This one is violating the AXI4-Stream specification. It is *NOT* allowed for Valid signals to depend on Ready (the other way round it is OK, so the issue does not affect the merger). This rule was defined in the AXI-Specification to avoid deadlock situations where both sides wait for each other.

I also think we should allow different consumers to consume the content in different clock cycles. Again, this is given by the AXI-spec: Valid is NOT allowed to be removed until the transaction happened but Ready IS ALLOWED to be removed at any time. Hence slaves could assert their "Ready" with a fixed pattern (which some slaves really do). Se below example

```
                   __      __      __           
Slave 0 Ready   __|  |____|  |____|  |__
                       __      __      _
Slave 0 Ready   ______|  |____|  |____| 
```

To accept this situation, we need to store what slaves already consumed their part of the data. Hence, we need _Clk_ and _Rst_ inputs to this entity.

src/axi/vhdl/olo_axi_stream_merger.vhd
MiloradPetrovic1: Understood, will do.

src/axi/vhdl/olo_axi_stream_merger.vhd
MiloradPetrovic1: Yes, will do.

src/axi/vhdl/olo_axi_stream_merger.vhd
MiloradPetrovic1: Understood, will do.

src/axi/vhdl/olo_axi_stream_merger.vhd
MiloradPetrovic1: But in **VHDL-2008** it is possible to read output ports.

src/axi/vhdl/olo_axi_stream_splitter.vhd
MiloradPetrovic1: Yes, will do.

src/axi/vhdl/olo_axi_stream_splitter.vhd
MiloradPetrovic1: Understood, will do.

src/axi/vhdl/olo_axi_stream_splitter.vhd
MiloradPetrovic1: Understood, will do.

src/axi/vhdl/olo_axi_stream_splitter.vhd
MiloradPetrovic1: Sorry, my bad, how embarrassing. I am well aware of how the AXI protocol works, but it seems that I was not paying attention.

I will make the required changes.

src/axi/vhdl/olo_axi_stream_merger.vhd
obruendl: Do all tools support it?

I am generally OK to go to VHDL-2008 but only for things all tools support and switching to VHDL-2008 means updating lots of tool import scripts and testing effort.

So although it might work, I think the lowest effort approach is to add the signal. But I can offer you to do the VHDL-2008 upgrade (and update the scripts as mentioned) in the next release, so in future you can use VHDL-2008. OK?

Next release will also contain a more thorrough synthesis test setup which synthesizes every entity - this will simplify the answer to my "do all tools support" it questions in similar cases.

MiloradPetrovic1: I cannot say if all tools support this feature, but AMD should be good. But I would definitely be interested in updating *olo* for VHDL-2008.

I really like to insist on VHDL-2008, and from my experience with AMD, they have decent support. But you obviously have much more experience with different vendors (like Gowin and Efinix - I never touched those).

For now, I will do as you say (add the buffer signal).

MiloradPetrovic1:
I will get back to you tomorrow  for this.

src/axi/vhdl/olo_axi_stream_merger.vhd
obruendl: Note on VHDL-2008:

I have set up the Synthesis CI environment. So I now can synthesize all open logic elements with all tools.

This has two effects:
1. I had to fix a few things that not all tools accepted (especially around tri-state ports) - these are non-reverse-compatible changes and hence the next release will be 4.0.0. As a result, I aim to execute all non-reverse-compatible changes known now, so we can avoid them afterwards for a while.
2. I actually can check now if the VHDL 2008 statements used are accepted by all tools. So we can safely switch to VHDL 2008.

Hence, you can expect to find VHDL 2008 on develop in near future.

obruendl:
Any update on this one? Please let me know when I shall review.

obruendl:
I see activity. Is the work still in progress or do you wait for a review from my side?

MiloradPetrovic1:
@obruendl Work in progress. I will let you know when it is time for a review.

Issue #162, opened 2025-06-11
[FEATURE REQUEST] WRR (Weighted Round Robin) Arbiter


rbrglez:
I propose that WRR arbiter is added to open-logic.

It could Look something like this:

```vhdl
entity olo_base_arb_wrr is
    generic (
        Grant_Width_g  : positive;
        Weight_Width_g : positive := 4
    );
    port (
        Clk : in std_logic;
        Rst : in std_logic;

        ------------------------------------------------------------------------
        -- Weight Interface
        ------------------------------------------------------------------------
        In_WeightValid : in  std_logic;
        In_WeightReady : out std_logic;
        In_Weight      : in  std_logic_vector(Weight_Width_g * Grant_Width_g - 1 downto 0);

        ------------------------------------------------------------------------
        -- Arbitration Interface
        ------------------------------------------------------------------------
        In_Req    : in  std_logic_vector(Grant_Width_g-1 downto 0);
        Out_Grant : out std_logic_vector(Grant_Width_g-1 downto 0);
        Out_Ready : in  std_logic;
        Out_Valid : out std_logic
    );
end entity;
```

`In_Weight` would be concatenated vector of weights for each 'Requestor'.

Example: 
Grant_Width_g = 3,
Weight_Width_g = 2,

Weight1 = 3 = "11"
Weight2 = 1 = "01"
Weight3 = 2 = "10"

In_Weight = Weight3 &  Weight2  & Weight1 = "110110"
NOTE: '&' is used as concatenation operator.

This would then Produce following Out_Grant,  Supposing In_Req = "111" and Out_Ready = '1'
G3, G3, G2, G1, G1, G1, G3, G3, G2, G1, G1, G1, ....
G3 = "100", G2 = "010", G1 = "001"

@obruendl, what do you think?

obruendl:
PR exists, hence - let's discuss in the PR instead of the issue. I close the issue therefore.

PR #163, opened 2025-06-12
WRR (Weighted Round Robin) Arbiter

rbrglez:
#162

Only the entity has been implemented so far.
TODO: Add documentation and VUnit testbenches.

@obruendl, could you please review my implementation of the WRR arbiter?
The design is functionally correct, but I believe it could be written more cleanly and efficiently.
I’m particularly concerned about the long combinatorial paths, what’s your take on that?


src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
+---------------------------------------------------------------------------------------------------
+-- Copyright (c) 2018 by Paul Scherrer Institute, Switzerland
+-- Copyright (c) 2024 by Oliver Bründler
+-- All rights reserved.
+-- Authors: Oliver Bruendler
```
MiloradPetrovic1: I assume you want your name here?

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+---------------------------------------------------------------------------------------------------
+-- TODO
+
+---------------------------------------------------------------------------------------------------
+-- Libraries
+---------------------------------------------------------------------------------------------------
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+use ieee.math_real.all;
```
MiloradPetrovic1: If you look at the other modules in this project, the "use ieee.---" lines are intendet so that all "ieee" are in line with the "ieee" from the "library ieee" line. So like this:
library ieee;
&nbsp;&nbsp;&nbsp;&nbsp;use ieee.std_logic_1164.all;
&nbsp;&nbsp;&nbsp;&nbsp;use ieee.numeric_std.all;
&nbsp;&nbsp;&nbsp;&nbsp;use ieee.math_real.all;

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+
+library work;
+use work.olo_base_pkg_math.all;
+use work.olo_base_pkg_logic.all;
```
MiloradPetrovic1: Same as comment above.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+    generic (
+        GrantWidth_g  : positive;
+        WeightWidth_g : positive
+    );
+    port (
+        Clk        : in  std_logic;
+        Rst        : in  std_logic;
+        In_Weights : in  std_logic_vector(WeightWidth_g * GrantWidth_g - 1 downto 0);
+        In_Req     : in  std_logic_vector(GrantWidth_g-1 downto 0);
+        Out_Grant  : out std_logic_vector(GrantWidth_g-1 downto 0);
```
MiloradPetrovic1:
```suggestion
        In_Req     : in  std_logic_vector(GrantWidth_g - 1 downto 0);
        Out_Grant  : out std_logic_vector(GrantWidth_g - 1 downto 0);
```

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+    );
+end entity;
+
+architecture rtl of olo_base_arb_wrr is
+
+    ----------------------------------------------------------------------------
+    -- Functions
+    ----------------------------------------------------------------------------
+    -- Returns the index of the first high ('1') bit in the vector
+    function getFirstHighBitIndex(vec : std_logic_vector) return integer is
```
MiloradPetrovic1: I assume a function like this would belong in *olo_base_pkg_logic.vhd*.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+        WeightCnt    : unsigned(WeightWidth_g - 1 downto 0);
+        WeightActive : unsigned(WeightWidth_g - 1 downto 0);
+        GrantIdx     : natural;
+    end record;
+
+    signal r      : TwoProcess_t;
+    signal r_next : TwoProcess_t;
+
+    -- Component connection signals
+    Signal RequestWeightsMasked     : std_logic_vector(In_Req'range);
```
MiloradPetrovic1:
```suggestion
    signal RequestWeightsMasked     : std_logic_vector(In_Req'range);
```

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+            variable v       : TwoProcess_t;
+            variable Grant_v : std_logic_vector(Out_Grant'range);
+        begin
+            -- hold variables stable
+            v := r;
+
+            -- Mask Requests with a weight of zero
+            RequestWeightsMasked <= In_Req and generateRequestWeightsMask(In_Weights, WeightWidth_g, GrantWidth_g);
+
+            -- Round Robing Logic
```
MiloradPetrovic1:
```suggestion
            -- Round Robin Logic
```

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+            port map (
+                Clk       => Clk,
+                Rst       => Rst,
+                In_Req    => RequestWeightsMasked,
+                Out_Grant => GrantRoundRobingUnmasked
+            );
+
+    end generate;
+
+end architecture;
```
MiloradPetrovic1, line 190: It is nice to have one empty line at the end of the file.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+            if rising_edge(Clk) then
+                r <= r_next;
+                if Rst = '1' then
+                    r.RoundRobingMask <= (others => '0');
+
+                    r.WeightCnt <= (others => '0');
+
+                    r.GrantIdx <= 0;
+
+                    r.WeightActive <= (others => '0');
```
MiloradPetrovic1:
```suggestion
                    r.RoundRobingMask <= (others => '0');
                    r.WeightCnt       <= (others => '0');
                    r.GrantIdx        <= 0;
                    r.WeightActive    <= (others => '0');
```

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+            end if;
+        end loop;
+
+        return requestWeightsMask;
+    end function;
+
+    -- Two Process Method
+    type TwoProcess_t is record
+        RoundRobingMask : std_logic_vector(In_Req'range);
+        --
```
MiloradPetrovic1: No sure what you mean by this empty comment?

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,198 @@
...
+
+            -- Apply to record
+            r_next <= v;
+
+        end process;
+
+        -- *** Sequential Process ***
+        p_seq : process (Clk) is
+        begin
+            if rising_edge(Clk) then
```
MiloradPetrovic1: What are your thoughts on using

`wait until rising_edge(Clk);`

instead of

`if rising_edge(Clk) then
...
end if;`

?

I prefer the former since:
- Reduces one level of indentation.
- No need for sensitivity list.
- Makes sure you have made a synchronous process (less error prone).
- Looks more *modern* :smile:

src/base/vhdl/olo_base_arb_wrr.vhd
obruendl: Currently my name should go there so I could change the license if needed (e.g. switch to mozilla or dual licensing to finance the project).

I am aware that this is not optimal and I want to give contributors the credits they deserve. The discussion here was the trigger for me to start setting up a contributor license agreement. I expect this to take a few days - and after this your name can go there.

src/base/vhdl/olo_base_arb_wrr.vhd
obruendl: Are you sure this works in all tools? Atleast chat gpt says yosys and synplify do not accept it. Hence my vote here would go for the more conservative "if" version. The downside is marginal - compared to the possible downside of some tool not accepting it.

If you feel it to be really advantageous, please try out at least in vivado, quartus, efinity, libero (synplify), gowin and yosys (coming soon for cologne chip) and post the results here.

I s  not against modern coding - just burnt by two decades full of findings about not every tool accepting allvalid hdl code.

obruendl: I have an update on this one:

* I tried in all tools and they accept it
* I discussed the topic with a few experienced FPGA designers in my network
* **My decision is**: We stay with the `if rising_edge(clk)` syntax for now. Reasons are:
  * This is the most common syntax, any other syntax may confuse readers.
  * The _if_ syntax works in all cases (also with asynchronous resets). We do not generally use asynchronous resets in open-logic - but still I prefer the generally valid syntax over having two syntaxes for different situations.
  * Althogh todays tools accept the `wait until`syntax, I do not know about older versions and also not about what non-officially supported tools users compile open-logic with. Hence there still is a risk to negatively impact happy users.

The change would not bring any benefits for the users (in contrary - it still has some risk). At maximum it would be in favor of open-logic developers (and I am unsure here as well, it's a matter of personal taste to some extent). My mindset is centered around the user - and from that perspective the situation is pretty clear IMO.

Note: This is my decision **today**. If at some point the `wait until` synthax becomes so popular that people are surprised to see the `if` synthax we should change it. But this IMO is not the case yet today.

src/base/vhdl/olo_base_arb_wrr.vhd
rbrglez: I have added my name under Authors, but left only your name under Copyright

src/base/vhdl/olo_base_arb_wrr.vhd
rbrglez: Yes, I think these functions could go to olo_base_pkg_logic. @obruendl what do you think?

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,204 @@
...
+-- Note: The link points to the documentation of the latest release. If you
+--       use an older version, the documentation might not match the code.
+
+---------------------------------------------------------------------------------------------------
+-- Libraries
+---------------------------------------------------------------------------------------------------
+library ieee;
+    use ieee.std_logic_1164.all;
+    use ieee.numeric_std.all;
+    use ieee.math_real.all;
```
obruendl: Is there any _real_ math used anywhere? I didn't find anything so probably this can be removed.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,204 @@
+---------------------------------------------------------------------------------------------------
+-- Copyright (c) 2018 by Paul Scherrer Institute, Switzerland
+-- Copyright (c) 2024 by Oliver Bründler
```
obruendl: You can put your name here. With signgature of the CLA this is fine and I still have the rights I need. Take the credits you deserve.

src/base/vhdl/olo_base_arb_wrr.vhd
obruendl: Yes, it should go to _olo_base_pkg_logic_ 

Please document the behavior for "to" and "downto" vectors cleanly. What is the _first_ bit in the two cases could be understood differently.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,204 @@
...
+                requestWeightsMask(i) := '0';
+            end if;
+        end loop;
+
+        return requestWeightsMask;
+    end function;
+
+    -- Two Process Method
+    type TwoProcess_t is record
+        RoundRobingMask : std_logic_vector(In_Req'range);
```
obruendl: Typo, should be "RoundRobin" (not Robin**g**)

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,204 @@
...
+    signal GrantRoundRobingMasked   : std_logic_vector(Out_Grant'range);
+    signal GrantRoundRobingUnmasked : std_logic_vector(Out_Grant'range);
+
+begin
+
+    -- Only generate code for non-zero sized arbiters to avoid illegal range delcarations
+    g_non_zero : if GrantWidth_g > 0 generate
+
+        -- *** Combinatorial Process ***
+        p_comb : process (r, In_Req, Out_Ready, In_Weights, GrantRoundRobingMasked, GrantRoundRobingUnmasked, RequestWeightsMasked) is
```
obruendl: Open Logic is now VHDL-2008 (will be officially released this week, already is the case on _develop_ where the PR goes). So I suggest you use `process (all) is` - no need to maintain sensitivity lists for combinatorial processes manually anymore.

src/base/vhdl/olo_base_arb_wrr.vhd
obruendl: I noticed it is not a typo but consistently wrong. Ideally you fix evrywhere.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,204 @@
...
+            RequestRoundRobingMasked <= RequestWeightsMasked and r.RoundRobingMask;
+
+            -- Generate Grant
+            if unsigned(GrantRoundRobingMasked) = 0 then
+                Grant_v := GrantRoundRobingUnmasked;
+            else
+                Grant_v := GrantRoundRobingMasked;
+            end if;
+
+            -- Get Weight of a currently active Grant
```
obruendl: Would it make sense to add a pipeline register somewhere on the path?

To me it looks like this will produce a quite significant combinatorial path - but I might be wrong. How many logi levels does this produce for a _GrantWidth_g=8 & WeightWidth_g=4_ implementation?

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,204 @@
...
+                v.WeightActive := unsigned(In_Weights((v.GrantIdx + 1) * WeightWidth_g - 1 downto v.GrantIdx * WeightWidth_g));
+            else
+                v.WeightActive := (others => '0');
+            end if;
+
+            -- Update RoundRobingMask
+            if (unsigned(Grant_v) /= 0) and (Out_Ready = '1') then
+                v.WeightCnt := r.WeightCnt + 1;
+
+                if not (r.WeightCnt < v.WeightActive - 1) then
```
obruendl: Why `not (A < B)` instead of ` A >= B`?

sim/test_configs/olo_base.py
```diff
@@ -213,6 +213,14 @@ def add_configs(olo_tb):
     arb_rr_tb = 'olo_base_arb_rr_tb'
     #Only one config required, hence no "add_config" looping
 
+    ### olo_base_arb_wrr ###
+    arb_wrr_tb = 'olo_base_arb_wrr_tb'
+    tb = olo_tb.test_bench(arb_wrr_tb)
+    for Grant in [3, 5, 8]:
```
obruendl: Could we check the corner case 1?

I tend to keep the nubmer of combinations low to avoid very long simulation times (not an issue with GHDL  and NVC but with the free modelsim it is an issue because it cannot be ran on multiple threads in parallel) . Hence: Is there a reson for checkign 5 and 8 or would one of the two do?

sim/test_configs/olo_base.py
```diff
@@ -213,6 +213,14 @@ def add_configs(olo_tb):
...
+        for Weight in [2, 3, 4]:
```
obruendl: Could we check the corner case 1?

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
```diff
@@ -0,0 +1,624 @@
+---------------------------------------------------------------------------------------------------
+-- Copyright (c) 2018 by Paul Scherrer Institute, Switzerland
```
obruendl: Did you copy any significant code from an existing TB with this copyright? If not, it does not have to be here.

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
```diff
@@ -0,0 +1,624 @@
...
+-- Copyright (c) 2025 by Oliver Bründler
```
obruendl: Same comment as for the implementation: Your name can go here.

doc/base/olo_base_arb_wrr.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Status Information
+
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/coverage/olo_base_arb_wrr.json?cacheSeconds=0)
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/branches/olo_base_arb_wrr.json?cacheSeconds=0)
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/issues/olo_base_arb_wrr.json?cacheSeconds=0)
+
+VHDL Source: [olo_base_arb_wrr](../../src/base/vhdl/olo_base_arb_wrr.vhd)
+
+## Description
```
obruendl, line 15: Could you add a waveform to the doc?

I usually draw them with wavedrom.

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
```diff
@@ -0,0 +1,624 @@
...
+                check_equal(Out_Grant, 2#00001#, "Out_Grant 16 not kept");
+                check_equal(Out_Valid, '1', "Valid low unexpectedly");
+                Out_Ready <= '0';
+                In_Req    <= "00000";
+                wait for Clk_Period_c * 10;
+            end if;
+
+            --------------------------------------------------------------------
+            --------------------------------------------------------------------
+            if run("RRLikeWeights_ExampleFromDoc") and GrantWidth_g = 5 then
```
obruendl: Says "example from the doc" but there is no example in the doc. I asked for one in the MD file, so once you draw it update the TB accordingly.

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
```diff
@@ -0,0 +1,624 @@
...
+                            LastCheckedGrant => LastCheckedGrant_v
+                        );
+                    end loop;
+                end loop;
+                wait for Clk_Period_c * 10;
+            end if;
+
+            --------------------------------------------------------------------
+            --------------------------------------------------------------------
+            if run("RandomNonZeroWeights_AllBits") then
```
obruendl: The name is not correct I think. To my understanding RandInt(1, 2**WeightWidth_g-1) will never produce a zero weight. Or am I wrong here?

sim/test_configs/olo_base.py
```diff
@@ -213,6 +213,14 @@ def add_configs(olo_tb):
...
+            for Seed in [42, 66, 11]:
```
obruendl: I suggest removing iterating through seeds. Yes, it is good from a coverage perspective but simulation runtime also is an issue as explained before.

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
```diff
@@ -0,0 +1,624 @@
...
+    use OSVVM.RandomPkg.all ;
+
+library olo;
+    use olo.olo_base_pkg_math.all;
+    use olo.olo_base_pkg_logic.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity Declaration
+---------------------------------------------------------------------------------------------------
+entity olo_base_arb_wrr_tb is
```
obruendl, line 28: Could you add

`-- vunit: run_all_in_same_sim`

This makes VUnit executing all test-cases in one single simulation which saves lots of time (especially with modelsim).

obruendl: I tried it out and it does not work yet - somehow the test-cases do impact each other. Probably you have to investigate.

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
```diff
@@ -0,0 +1,624 @@
...
+                for i in 0 to GrantWidth_g - 1 loop
+                    In_Weights((i + 1) * WeightWidth_g - 1 downto i * WeightWidth_g) <=
+                        toUslv(1, WeightWidth_g);
+                end loop;
+                -- Always Rdy
+                Out_Ready <= '1';
+                wait until rising_edge(Clk);
+                In_Req <= "10000";
+                wait for Tpd_c;
+                check_equal(Out_Grant, 2#10000#, "Out_Grant 3 Wrong");
```
obruendl: I know, you copied the style from the arb_rr TB. Unfortunately this is a rather bad example (it's one that I took over from the PSI libraries so I did simply leave it very close to how it was).

If I'd write it from scratch I'd probably either use AXI-Stream VCs instead of the long manually written sequenes or at least write a procedure to which I can pass the sequence of Grant vectors over a number of cycles (as array of std_logic_vector).

I leave the decision to you if you want to clean up the TB or if you want to leave it as it is. Because the style is copied from arb_rr I would find it justifiable (not optimal) in this case to leave it.

obruendl:
General question:

Couldn't the arb_wrr be created based on (instantiating it) arb_rr instead of duplicating quite a bit of its code?

Generall idea:
1. Mask all In_Req bitw with weight zero
2. Pass this masked request vector to arb_rr
3. On the output of the arb_rr only assert Rrady after the the number of clock cycles that correspond to the grant index

src/base/vhdl/olo_base_arb_wrr.vhd
rbrglez: removed

src/base/vhdl/olo_base_arb_wrr.vhd
rbrglez: done

src/base/vhdl/olo_base_arb_wrr.vhd
rbrglez: done

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
rbrglez: obsolete comment. testbench went through major rewrite.

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
rbrglez: Added run_all_in_same_sim

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
rbrglez: Rewritten arb_wrr testbench using AxiStream VC

rbrglez:
## Recap of what was done:

- Major rewrite of olo_base_arb_wrr_tb
- Added `get*SetBitIndex` functions to olo_base_pkg_logic (also added documentation and vunit tb)
- Fixed some minor things raised in this review (removed math library, used all keyword, updated some comments, ....)

## Still on my TODO list:

- Update Documentation for WRR entity
- Investigate if the WRR entity can be restructured to reuse the RR entity internally (refer to @obruendl’s general question)
- Squash commits

rbrglez:
@obruendl,
Can the generic standard logic array `StlvArray_t` be used on ports, or is it intended for internal use only?
What is the policy regarding its usage?

rbrglez:
> General question:
> 
> Couldn't the arb_wrr be created based on (instantiating it) arb_rr instead of duplicating quite a bit of its code?
> 
> Generall idea:
> 
>     1. Mask all In_Req bitw with weight zero
> 
>     2. Pass this masked request vector to arb_rr
> 
>     3. On the output of the arb_rr only assert Rrady after the the number of clock cycles that correspond to the grant index

I have rewritten arb_wrr, so that it uses arb_rr as an instance as you suggested. I have implemented it twice (with latency and without latency)

Can you review the implementation?
Both implementations pass all the vunit tests (I had to make a slight change in one of the tests, so that latency implementation passed)

src/base/vhdl/olo_base_arb_wrr.vhd
obruendl, line 1: I think the code could be reduced quite a bit by putting the request masking andcthe olo_base_arb_rr instantiation into the olo_base_arb_wrr directly and only have two separate if-generate statements for the output side logic.

Is there anything that prevents this approach?

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,455 @@
...
+        variable v : TwoProcess_t;
+    begin
+        -- hold variables stable
+        v := r;
+
+        v.RR_GrantReady := '0';
+
+        -- Get the Weight value for the currently active Grant
+        if (RR_GrantValid = '1') then
+            v.GrantIdx := getLeadingSetBitIndex(RR_Grant);
```
obruendl: If I didn't miss anything (working on a tablet is not like on a PC...)  GrantIdx is alocal variable in this loop only. Correct?

If sp, better define it as variable for this process only insted of as part of the two process record. The two process record should only contain things that synthesize to registers.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,455 @@
...
+            -- Extract the corresponding weight using the GrantIdx
+            v.Weight :=
+                unsigned(In_Weights((v.GrantIdx + 1) * WeightWidth_g - 1 downto v.GrantIdx * WeightWidth_g));
+        end if;
+
+        if (v.GrantValid = '1' and Out_Ready = '1') then
```
obruendl: I think this should be r.GrantValid.

Whenever possible use read from r.xxx to emphasize that this is a register without combinatorial logic. Only reading v.xxx in rare corner cases also helps to avoid creating long combinatorial paths.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,455 @@
...
+            -- Increment the weight counter on each successful AXI handshake
+            v.WeightCnt := r.WeightCnt + 1;
+
+            -- If the same grant has been used for 'Weight' handshakes,
+            -- assert RR_GrantReady to request the next grant and reset the counter
+            if (v.WeightCnt >= v.Weight) then
+                --v.GrantValid    := '0';
```
obruendl: Dead code? If so, remove

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,455 @@
...
+    use work.olo_base_pkg_logic.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity Declaration
+---------------------------------------------------------------------------------------------------
+entity olo_base_arb_wrr is
+    generic (
+        GrantWidth_g  : positive;
+        WeightWidth_g : positive;
+        Latency_g     : boolean
```
obruendl: Could we make it natural range 0 to 1? This would match the existing Latenc_g for the priority arbiter

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,455 @@
...
+            --------------------------------------------------------------------
+            when GetRrGrant_s =>
+                v.RR_GrantReady := '1';
+
+                if (RR_GrantValid = '1' and r.RR_GrantReady = '1') then
+                    v.RR_GrantReady := '0';
+                    v.Grant         := RR_Grant;
+
+                    if (unsigned(RR_Grant) /= 0) then
+                        v.GrantIdx := getLeadingSetBitIndex(RR_Grant);
```
obruendl: Same comment as further up. Should be a local variable I think.

src/base/vhdl/olo_base_arb_wrr.vhd
obruendl: Probably my comment was unclear here - I meant avoiding the component instantiation (and write the code into generates). I pushed an according change and the code size is reduced quite a bit. please review.

obruendl:
Lets assume `GrantWidth_g=2`, `WeightWidth_g=3` and `Latency_g=1`

Now let's assume `In_Req="11"` and all weights = 3 . After one clock cycle we will get `Out_Grant="10"` and `Out_Valid=1` at the output and this is immediately accepted by `Our_Ready=1`. If the user logic now detects that on channel 1 which just got the grant it does not need more grants, it will do `In_Req="01"`. However, I think (correct me if wrong), we would still see `Out_Grant="10"` and `Out_Valid=1`  for one more clock cycle, until the change in `In_Req` is reflected due to the register stage and in the cycle after we get `Our_Valid='0'`. From an implementation perspective this makes sense.

But from a user perspective, we have the following issues:
- We have seen a grant on channel 1 plust the Valid and the grant is removed without us having applied `Our_Ready='1'`. This is somewhat unlogigcal.
- Switching a `Valid` Signal from 1 to 0 without a `Ready=1` is against the AXI-Stream definitions

As mentioned, from an implementation perspective this makes perfect sense and I could not imagine how to avoid this with the current interface. Backpressure (Ready) seems to be difficult in this case.

I thought a bit about it and I could not find a very straightforward solution. What I could think of, woudl be:
a) We remove `Latency_g`. I know I suggested it, but if it leads to a difficult to understand interface and wen cannot fix this, we probably better offer less features in a understandable way.
b) Instead of using `Out_Ready` we could add `In_Valid` - in this case the user could request each arbitration round explicitly and get the result one clock cycle later. He can then do whatever he wants (e.g. send a packet from the channel that got the grant) and the nrequest the next arbitration round by `In_Valid=1`. 

Could you check if my interpretations are correct? If not, you best quickly point me to the simulation case that covers the situation.

Should I be correct, please let me know what you think of it. I think this requires discussion before implementation. It's a case where I feel that the current situaiton is not optimal from a users perspective but I am not 100% sure about how to solve it best.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,248 @@
...
+                            GrantIdx_v := getLeadingSetBitIndex(RrGrant);
+                            v.Weight   := unsigned(In_Weights((GrantIdx_v + 1) * WeightWidth_g - 1 downto GrantIdx_v * WeightWidth_g));
+                            v.State    := SendGrant_s;
+                        end if;
+                    end if;
+
+                --------------------------------------------------------------------
+                when SendGrant_s =>
+                    -- Check if grant can still be sent
+                    if (ReqMasked(GrantIdx_v) = '1' and r.WeightCnt <= r.Weight - 1) then
```
obruendl: I just noticed that you youse GrantIdx_v here. I overlooked that before - in this case it shoudl go into the two process record (otherwise we are building a latch).

obruendl (changes requested):

rbrglez:
I've prepared a set of example waveforms to illustrate how the WRR arbiter would behave if we remove the `Out_Ready` signal and instead use an `In_Valid` signal.

All the waveforms assume a 2-bit `In_Req`, typically set to `"11"`, and modified in certain cases to `"01"` or `"00"`. The `In_Weights` remain constant across all cases: `(2, 1)`.

@obruendl , can you provide some feedback on generated waveforms?

## Slow Simple
```
{signal: [
  {name: 'Clk',        wave: 'p..............'                                         },
  {name: 'In_Valid',   wave: '0.101010101010.'                                         },
  {name: 'In_Weights', wave: 'x.2............', data: ['(2,1)']                        },
  {name: 'In_Req',     wave: 'x.2............', data: ['"11"']                         },
  {                                                                                    },
  {name: 'Out_Valid',  wave: '0..101010101010'                                         },
  {name: 'Out_Grant',  wave: 'x..3...4.3.4...', data: ['"10"', '"01"', '"10"', '"01"'] },
]}
```

slow_simple

## Slow Request Changed
```
{signal: [
  {name: 'Clk',        wave: 'p.............'                                         },
  {name: 'In_Valid',   wave: '0101010101010.'                                         },
  {name: 'In_Weights', wave: 'x2............', data: ['(2,1)']                        },
  {name: 'In_Req',     wave: 'x3.4...3......', data: ['"11"', '"01"', '"11"']         },
  {                                                                                   },
  {name: 'Out_Valid',  wave: '0.101010101010'                                         },
  {name: 'Out_Grant',  wave: 'x.3.4...3...4.', data: ['"10"', '"01"', '"10"', '"01"'] },
]}
```
slow_req_changed

## Slow Request Changed to Zero
```
{signal: [
  {name: 'Clk',        wave: 'p.............'                                         },
  {name: 'In_Valid',   wave: '0101010101010.'                                         },
  {name: 'In_Weights', wave: 'x3............', data: ['(2,1)']                        },
  {name: 'In_Req',     wave: 'x3.4...3......', data: ['"11"', '"00"', '"11"']         },
  {                                                                                   },
  {name: 'Out_Valid',  wave: '0.10....101010'                                         },
  {name: 'Out_Grant',  wave: 'x.5.8...5...6.', data: ['"10"', '"00"', '"10"', '"01"'] },
]}
```
slow_req_zero_changed

## Fast Simple
```
{signal: [
  {name: 'Clk',        wave: 'p..............'                                                                         },
  {name: 'In_Valid',   wave: '0.1............'                                                                         },
  {name: 'In_Weights', wave: 'x.2............', data: ['(2,1)']                                                        },
  {name: 'In_Req',     wave: 'x.2............', data: ['"11"']                                                         },
  {                                                                                                                    },
  {name: 'Out_Valid',  wave: '0..1...........'                                                                         },
  {name: 'Out_Grant',  wave: 'x..3.43.43.43.4', data: ['"10"', '"01"', '"10"', '"01"', '"10"', '"01"', '"10"', '"01"'] },
]}
```
fast_simple

## Fast Request Changed
```
{signal: [
  {name: 'Clk',        wave: 'p...........'                                                         },
  {name: 'In_Valid',   wave: '0.1.........'                                                         },
  {name: 'In_Weights', wave: 'x.3.........', data: ['(2,1)']                                        },
  {name: 'In_Req',     wave: 'x.3..4.3....', data: ['"11"', '"01"', '"11"']                         },
  {                                                                                                 },
  {name: 'Out_Valid',  wave: '0..1........'                                                         },
  {name: 'Out_Grant',  wave: 'x..5.66.5.65', data: ['"10"', '"01"', '"01"', '"10"', '"01"', '"10"'] },
]}
```
fast_req_changed

## Fast Request Changed to Zero
```
{signal: [
  {name: 'Clk',        wave: 'p...........'                                                         },
  {name: 'In_Valid',   wave: '0.1.........'                                                         },
  {name: 'In_Weights', wave: 'x.3.........', data: ['(2,1)']                                        },
  {name: 'In_Req',     wave: 'x.3..4.3....', data: ['"11"', '"00"', '"11"']                         },
  {                                                                                                 },
  {name: 'Out_Valid',  wave: '0..1..0.1...'                                                         },
  {name: 'Out_Grant',  wave: 'x..5.68.5.65', data: ['"10"', '"01"', '"00"', '"10"', '"01"', '"10"'] },
]}
```
fast_req_zero_changed

obruendl:
@rbrglez Yes, that's exacrly what I imagined for Latency_g=1.

For Latency_g=0 it'd be the same, just all the Out_* shifted by one cycle to the left.

This will also simplify the TB a lot: you now can just let axi-steam-master VC inject such sequences (with random stall) and let axi-stream-slave check the output (without stall).

rbrglez:
@obruendl , I've updated the WRR arbiter. When you have some time, could you please review the changes? If no major revisions are needed, I’ll start writing the documentation after your review.

Here are snapshots of simulation:

## Latency_g = 0
Screenshot from 2025-07-06 22-31-50

## Latency_g = 1
Screenshot from 2025-07-06 22-33-16

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,258 @@
...
+                        end if;
+
+                    elsif (RrGrantValid = '0') then
+                        v.ReqInValid := '0';
+                    end if;
+
+                ----------------------------------------------------------------
+                when SendGrant_s =>
+                    -- Check if grant can still be sent
+                    if (ReqMasked(GrantIdx_v) = '1' and r.WeightCnt <= r.Weight - 1) then
```
obruendl: You use GrantIdx_v here which was set in the cycle before (in GetRdGrant_s). This creates a latch, which  iolates synchronous design principles.

test/base/olo_base_pkg_logic/olo_base_pkg_logic_tb.vhd
obruendl, line 1: .
Can wereduce tb code here by just using a representative set of literal values like it is the case for the other functions? With the currently proposed chsnge the tb becomes verx unbalanced and for my feelings at some point more tests just add more maintenance effort at relatively little benefit.

Alternatively - if you want to keep the full tests - move them to a separate tb and only add a representative set of literals to the current tb. This way the existing tb becomes balanced and we can still keep the additional tests.

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
obruendl, line 1: Is there any reason for not using an axi_stream_master vc for injecting weights and in_req? IMO this would simplify the TB a lot: you just inject weigh/req sequences and the expected grant sequences. This could be done non-blocking and with random stalling on the axi_stream_master.

obruendl (changes requested):

rbrglez:
While integrating the axis_stream_master Verification Component, I noticed that the WRR  logic didn't handle certain edge cases well. As a result, I had to make some significant changes, including adding an In_Ready port.

The code still needs some polishing, but hopefully this is the last major rewrite of the WRR.

@obruendl, please take a look at the updated testbench and let me know if you agree with the expected behavior. Specifically how In_Req and In_Weights map to the expected Out_Grant.

rbrglez:
Updated Testbench.
- Added RandomStall_g 
- Used consistent naming scheme for unit tests
- Removed obsolete stuff from testbench
- Beautified testbench

obruendl:
@rbrglez 

**Topic 1: Usage of StdlvArray_t on ports**

The general answer is "If this can be instantiated from Verilog with all the supported tools yes." (and in this case a sample of the verilog syntax should be provided). 

Cross language instantiation is somwhat undefined and I found out that all the tools have slightly different limitations. I do not know if they all would eat it. If you want to do it, provide a test-case (Verilog file instantiating the WRR arbiter with StdlvArray_t), test it on your side and I'd like to verify it on my side. 

I appreciate if you take this effort. But I also understand if you want to stay away from it and just continue with the flattened std_logic_vector.

**Topi 2: New Timing for Latency_g=1**

If I interpret the waveforms correctly, we now cannot accept one arbitration per clock cycle but only one every three clock cycles. Correct?

What is the reason for this? To my understanding it should be possible to do one arbitration per clock-cycle also in this case. Depending on the use-case this might be desireable IMO. Hence I'd like to understand where the limitation comes from (I probably overlook something).

rbrglez:
> **Topic 1: Usage of StdlvArray_t on ports**
> 
> The general answer is "If this can be instantiated from Verilog with all the supported tools yes." (and in this case a sample of the verilog syntax should be provided).
> 
> Cross language instantiation is somwhat undefined and I found out that all the tools have slightly different limitations. I do not know if they all would eat it. If you want to do it, provide a test-case (Verilog file instantiating the WRR arbiter with StdlvArray_t), test it on your side and I'd like to verify it on my side.
> 
> I appreciate if you take this effort. But I also understand if you want to stay away from it and just continue with the flattened std_logic_vector.


Maybe I'll do this in another PR.

If I understand correctly, then I can use StdlvArray_t internally in the WRR unit?

doc/base/olo_base_arb_wrr.md
obruendl, line 1: Documentation is not up to date (I noticed ports are not missing).

Please update. Please also add example waveforms for both, Latency_g=0 and Latency_g=1.

The markdown linter also reported errors. Please fix those.

doc/base/olo_base_pkg_logic.md
obruendl, line 1: The markdown linter reported errors. Please fix those.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,329 @@
...
+    use work.olo_base_pkg_logic.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity Declaration
+---------------------------------------------------------------------------------------------------
+entity olo_base_arb_wrr is
+    generic (
+        GrantWidth_g  : positive;
+        WeightWidth_g : positive;
+        Latency_g     : natural range 0 to 1
```
obruendl: Add a default value to Latency_g. I think 0 is a sane default here.

The other settings do not require a default - the user must decide on those.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,329 @@
...
+                -- Increment WeightCnt.
+                else
+                    v.WeightCnt := v.WeightCnt + 1;
+
+                end if;
+            end if;
+
+            -- Apply to record
+            r_next <= v;
+        end process;
```
obruendl: I noticed the two FSMs are _almost_ identical.

I think it probably would make more sense to only have one FSM and only handle the differences separately by clode like:

```
if (Latency_g=0 and unsigned(r.Grant)=0) or (Latency_g=1 and unsigned(v.Grant)=0)) then
```

I think this would reduce the risk for the two almost identical FSMs deviating in future accidentally.

src/base/vhdl/olo_base_arb_wrr.vhd
obruendl, line 1: The code still throws linter error. Please make it linter happy.

test/base/olo_base_arb_wrr/olo_base_arb_wrr_tb.vhd
obruendl, line 1: The TB still throws linter errors, please make it linter happy.

obruendl (changes requested):
The functionality seems now fine to me.

Please make everything linter-happy and complete the documentation (not updated, no waves yet). I fully understand you wanted to wait with this until I approve the code itself of course. Now its the time to pick those up again.

lint/config/vsg_config_overlay_vc.yml
obruendl, line 1: Why did you modify the linter configuration?

IMO adding a VHDL entity should not required any modification to the linting rules.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,329 @@
...
+    -- state record
+    type State_t is (
+        Idle_s,
+        SendGrant_s
+    );
+
+    -- Two Process Method
+    type TwoProcess_t is record
+        -- Round Robin
+        RrReq        : std_logic_vector(In_Req'range);
```
obruendl: r.RrReq does not seem to be read anywhere. Hence this is not a register and it should not be in the two process record.

I suggest assigning 

```
RrReq  <= In_Req and generateRequestWeightsMask(In_Weights, WeightWidth_g, GrantWidth_g);
```
separately and in the if-clause where `v.RrReq` is read currently reading the signal `RrReq`

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,329 @@
...
+                            r.WeightCnt = 0 or
+                            v.RrReq(r.GrantIdx) = '0' or
+                            r.WeightCnt >= v.Weight
+                        ) then
+                            -- Reset WeightCnt for new Grant
+                            v.WeightCnt    := (others => '0');
+                            v.RrGrantReady := '1';
+                            v.GrantIdx     := getLeadingSetBitIndex(RrGrant);
+                            -- Recalculate Weight as it may have changed because of new GrantIdx
+                            v.Weight := unsigned(In_Weights((v.GrantIdx + 1) * WeightWidth_g - 1 downto v.GrantIdx * WeightWidth_g));
```
obruendl: The Code is OK.

I just asked myself if it would make sense to add a functions for flattening arrays and unflatten them to olo_base_pkg_array. Something like:

```
array2Vector(a : StdlvArray_t) return std_logic_vector;
vector2Array(v : std_logic_vector, elementSize : integer) return StdlvArray_t;
```

You could then use a Weights Array internally. This would be more readable than the indexing.

What is your take on this?

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,329 @@
...
+                end if;
+
+                v.GrantValid := '1';
+                v.State      := SendGrant_s;
+
+            end if;
+
+            --------------------------------------------------------------------
+            -- Send The Grant (same for both GetRrGrant_s and SendGrant_s state)
+            if (v.State = SendGrant_s) then
```
obruendl: This is confusing to me. You read v.STate here, which was set a few lines above. To me it looks like we are not actually implementing an FSM here but state is always set to Idle_s in line 248 and we conditionally execute lines 278 to 297. But such conditional execution should be described by an "if" not by a state/FSM IMO to not confuse readers.

obruendl:
# Pipelined Implementation

I always had the feeling that it shoudl be possible to implement the WRR arbieter pipelined, so it can do one arbitration per clock cycles (also with _Latency_g=1_). I spent a bit of time and succeeded implementig this on the branch [here](https://github.com/open-logic/open-logic/tree/feature/pipelined-wrr).

This implementation also has much less code.

I ran synthesis with all supported tools and this looks OK as well. Also number of logic levels seems sane (4 levels for _GrantWidth_g=5_ and _WeightWidth_g=4_).

The only trade-off I had to do is making the _In_Weights_ check a bit different. The arbiter now switches to the next channel _after a grant was given and it is detected that the number of grants to the requester is above In_Weights provided with this sample_. This is described in the bullet list below

1. Let's assum In_Weights = (3, 3, 3) and In_Req = "111"
2. First grant is "100"
3. Second grant is "100"
4. Now let's assume In_Weights changes to (1, 3, 3). 
5. In the old implementation by @rbrglez the third grant would be "010", in my updated implementation still one more "100" grant is given (because the check for weights is executed after giving the grant)
6. The fourth grant is "010"
7. ...

rbrglez:
@obruendl please review.

I have updated Documentation, added a few more test cases used in documentation and made linter happy.
Check out commit, for inference tests. Add more inference tests if you want.

src/base/vhdl/olo_base_arb_wrr.vhd
```diff
@@ -0,0 +1,188 @@
...
+    signal r_next : TwoProcess_t;
+
+    -- Component connection signals
+    signal RrReq        : std_logic_vector(In_Req'range);
+    signal RrGrant      : std_logic_vector(Out_Grant'range);
+    signal RrGrantReady : std_logic;
+
+begin
+
+    RrReq <= In_Req and generateRequestWeightsMask(In_Weights, WeightWidth_g, GrantWidth_g);
```
obruendl: @rbrglez 

I noticed that here we use _In_Weights_ immediately (without the "one sample delay" we apply for the switchover mechanism).

IMO it would make sense to rename _In_Weights_ to just _Weights_ and say "this is a static configuration signal and changes to _Weights_ are applied to arbitration requests within less than 3 clock cycles" (similar to [olo_base_delay_cfg](https://github.com/open-logic/open-logic/blob/main/doc/base/olo_base_delay_cfg.md#description) delay changes). I'd assume _In_Weights_ to be static anyways.

Alternatively we can keep it handshaked. In this case we should change to `r.Weights` here but we also have to update the TB accordingly and. 

For me the "see _Weights_ as static control signal" approach would make more sense.

Let me know if you agree.

BTW: Do not change documentation. I am mostly done with my final review and enhancing doc a bit (that's how I came accross the issue)..

obruendl (approved):

rbrglez:
@obruendl 

Looks good, I have pushed some minor changes 
- fixed minor errors in documentation
- fixed JSON, from which PNG is generated (TODO: @obruendl, can you generate a new PNG?)
- updated comments in the WRR source code so that it aligns with the documentation
- Updated WRR TB so that ExampleFromDoc_* testcases align with the documentation

Thanks a lot for your detailed reviews and guidance throughout this pull request :smile:

PR #167, opened 2025-06-19
Change Language Standard to VHDL 2008

obruendl:
I changed the language standard to VHDL 2008 and updated all tool integration scripts to configure projects accordingly or import files explicitly with this standard (depending on the tool one or the other is applicable).

In this context I also brought back the unconstrained array types in _olo_base_pkg_array_ (because with VHDL 2008 all tools accept those).

I also simplified all sensitivity lists to `(all)` which is now possible.

I have synthesized the reference designs for all tools and they seem to accept it.

### DIFF
--- a/src/base/vhdl/olo_base_pkg_array.vhd
+++ b/src/base/vhdl/olo_base_pkg_array.vhd
@@ -62,7 +62,6 @@ package olo_base_pkg_array is
     type StlvArray64_t  is array (natural range <>) of std_logic_vector(63 downto 0);
     type StlvArray512_t is array (natural range <>) of std_logic_vector(511 downto 0);
 
-
     type IntegerArray_t is array (natural range <>) of integer;
     type RealArray_t is array (natural range <>) of real;
     type BoolArray_t is array (natural range <>) of boolean;


src/axi/vhdl/olo_axi_lite_slave.vhd
```diff
@@ -230,7 +228,7 @@ begin
     end process;
 
     -- *** Sequential Process ***
-    p_seq : process (Clk) is
+    p_seq : process (all) is
```
rbrglez, line 231: In sequential processes, I typically include only the clock (clk) in the sensitivity list and avoid using the all keyword.

obruendl: I thought about this but I then decided that it is easiest to juse use "all" everywhere.

rbrglez (commented):
Review finished, PR can be merged

rbrglez (approved):
Approve

src/axi/vhdl/olo_axi_lite_slave.vhd
tasgomes: I find this a very bad idea. Please find below my reasoning:

1. This influences the simulation performance since this process will wake up constantly instead of waking up only due to a change in the clock. 
2. This could also introduce unnecessary mismatches between simulation and synthesis, who knows how the vendor tools parse code to understand this should be a sequential logic only clocked by the clock signal. 
3. Async vs Sync Resets... you could be potentially mixing async and sync resets.

Issue #170, opened 2025-06-19
Using `all` in sensitivity list of sequential process


tasgomes:
I commented on the PR but it seems it went unnoticed during merge. I think using `all` in the sensitivity list of sequential processes might be a bad idea.

Example:

https://github.com/open-logic/open-logic/blob/f2ad980317378bef25c7fce0496e48acdf4ab737/src/axi/vhdl/olo_axi_lite_slave.vhd#L231

Please find below my reasoning:

1. This influences the simulation performance since this process will wake up constantly due to activity on any of the signals being read in the process instead of waking up only due to a activity on the clock signal.
2. This could also introduce unnecessary mismatches between simulation and synthesis, who knows how the vendor tools parse code to understand whether this should be a sequential logic that is only clocked by the clock signal.
3. Async vs Sync Resets... you could be potentially mixing async and sync resets, since your process could now be sensitive to the reset in an asynchronous way.

obruendl:
@tasgomes 

I didn't see any significant impact to simulation performance and synthesis results are 100% the same. However, the statement seems true according to some recherche - so I will change it back to clk for sequential processes.

Thanks for the input.

PR #177, opened 2025-06-21
IMPROVEMENT: Do not use "all" for synchronous processes

obruendl:
refs #170

@tasgomes I implemented your suggestion to NOT use _all_ for synchronous processes. Please review.


tasgomes:
One is missing. This one is not a synchronous process but I think having only the `Clk` makes sense:

https://github.com/open-logic/open-logic/blob/issue/170/src/axi/vhdl/olo_axi_master_simple.vhd#L252C4-L252C5

This is the same as here, where you already use the `Clk` only:

https://github.com/open-logic/open-logic/blob/issue/170/src/base/vhdl/olo_base_flowctrl_handler.vhd#L100

obruendl:
@tasgomes Good spot!

I fixed it. Do you agree that I merge the PR now?

tasgomes:
Yes.

PR #189, opened 2025-06-29
Deprecate strength strippers (to01, to01X)

rbrglez:


src/base/vhdl/olo_base_pkg_logic.vhd
```diff
@@ -227,6 +232,7 @@ package body olo_base_pkg_logic is
 
     end function;
 
+    -- DEPRECATED: use to_X01 from std_logic_1164
```
obruendl: @rbrglez 

I think it would make sense to put a warning here that says "deprecated - use ... instead". Just a "report ... severity warning".

obruendl (changes requested):
Besides the warning that I'd suggest to add and the markdown linter issues the PR seems fine.

BTW:  The MD linter you can add to VSCode by installing the _markdownlint_ plugin (see contributors.md).

I also tested modelsim simulation (NVC is in the PR checks) and at least it did compile fine. I don't expect runtime errors - but I'll let it finish anyways (it's still running).

src/base/vhdl/olo_base_pkg_logic.vhd
rbrglez: Yeah, this is much better than just adding a comment :)
Done

src/base/vhdl/olo_base_pkg_logic.vhd
```diff
@@ -218,6 +218,9 @@ package body olo_base_pkg_logic is
     function to01X (inp : in std_logic) return std_logic is
     begin
 
+        report "DEPRECATED - use 'to_X01' from IEEE.STD_LOGIC_1164 instead."
```
obruendl: Can we make the mesage "DEPRECATRD - olo_base_pkg_logic.to01x() is deprecatex - use ..."?

I am not sure if all tools always report the file the warning came from, so webetter be clear. Thats the last little thing.

Then please squash the changes to a single commit (message "CLEANUP: ...") and I will accept the PR. Thanks.

obruendl (changes requested):

src/base/vhdl/olo_base_pkg_logic.vhd
rbrglez: done and done :)

Issue #202, opened 2025-08-08
Unsafe path on LUTRAM when using async FIFO or olo_base_ram_sdp.vhd


tasgomes:
I am using the `olo_base_fifo_async` entity in Vivado 2024.2 as follows:

```vhdl
    i_fifo : entity work.olo_base_fifo_async
    generic map(
        Width_g                         => 27,
        Depth_g                         => 64,
        AlmFullOn_g                     => false,
        AlmFullLevel_g                  => 0,
        AlmEmptyOn_g                    => false,
        AlmEmptyLevel_g                 => 0,
        RamStyle_g                      => "auto",
        RamBehavior_g                   => "RBW",
        ReadyRstState_g                 => '1',
        Optimization_g                  => "SPEED",
        SyncStages_g                    => 2
    )
```

Due to the small size of it, the RAM style ends up being a LUTRAM:

<img width="765" height="650" alt="Image" src="https://github.com/user-attachments/assets/38f039c6-d811-49f8-beb4-e44b19515d5c" />

In this scenario, I realized that the inter-clock path between these two clock domains was not being constrained with an exception, i.e., Vivado was analyzing them as a timed path. Depending on the clock source, their frequencies and phases, the timing on this path can fail.

I now added a scoped-constrained for this path (a merge-request will follow). 

However, as described in [UG906](https://docs.amd.com/r/en-US/ug906-vivado-design-analysis/LUTRAM-Read/Write-Potential-Collision), this path could still be unsafe if there is a read and write access to the same address at the same time. 

So unless there is logic that prevents any read/write collision I think it is unsafe to use this block (in LUTRAM mode). 

At the end, I am not sure if adding a scoped constraint is recommend as this will hide a potential dangerous path.

tasgomes:
An async FIFO never allows a valid read and write to the same address at the same time, by design. 

Hence, this path can be safely ignored from timing analysis using a scoped constraint when the entity `olo_base_ram_sdp.vhd` is used in the context of an async FIFO.

However, when this entity is used as standalone, it is probably not a good idea to apply automatic scoped constraints on this path. The designer might not be aware that this path could be unsafe if there is a read and write access to the same address at the same time. It is better to let Vivado warn the user about it.

Hence, I propose we only add a scope constraint for the async FIFO module.

obruendl:
PR merged, hence I close this issue. @tasgomes Feel free to reopen if you disagree

PR #203, opened 2025-08-08
Add scoped constraint for olo_base_ram_sdp.tcl

tasgomes:
@obruendl as mentioned in this issue https://github.com/open-logic/open-logic/issues/202, this scoped constraint mark a CDC path as un-timed when the RAM type is LUTRAM.

However before we merge this we need to address 3 topics:

**Topic 1**
As mentioned in the issue, I think this path is not safe if read and write operations to the same address are done, so **I would not add this scoped constraint and until that is cleared out.**

**Topic 2**
The scoped constraints are currently being added as "XDC" files and not TCL. The scoped constrained implemented in this pull request needs the file type to be TCL as it executes an if-statement.

image

In the past you were adding this as TCL, note how the icon is different:

image

What do we need to do to add them as TCL?

**Topic 3**

I tried to force the Async FIFO to be implemented as BLOCKRAM so I could test my if-statement above:
```vhdl
    i_fifo : entity work.olo_base_fifo_async
    generic map(
        Width_g                         => fifo_din'length,
        Depth_g                         => 4096*4, -- LARGE
        AlmFullOn_g                     => false,
        AlmFullLevel_g                  => 0,
        AlmEmptyOn_g                    => false,
        AlmEmptyLevel_g                 => 0,
        RamStyle_g                      => "block", -- BLOCK!!!!
        RamBehavior_g                   => "RBW",
        ReadyRstState_g                 => '1',
        Optimization_g                  => "SPEED",
        SyncStages_g                    => 2
    )
```

However, Vivado refuses and still implements it as LUTRAM:

image

In your synthesis/PAR tests, did you ever check if your ASYNC FIFO can be mapped to BRAMs?

### DIFF
--- a/.github/workflows/rd_build.yml
+++ b/.github/workflows/rd_build.yml
@@ -105,13 +105,11 @@ jobs:
           gw_sh scripted_build_sv_sh.tcl
       # Libero
       - name: Libero VHDL
-        if: false #Libero currently fails on AWS - issue under innvestigation. RD build is checked manually.
         run: |
           source $LOCAL_TOOLS
           cd ./doc/tutorials/LiberoTutorial/Files
           libero script:scripted_build.tcl
       - name: Libero Verilog
-        if: false #Libero currently fails on AWS - issue under innvestigation. RD build is checked manually.
         run: |
           source $LOCAL_TOOLS
           cd ./doc/tutorials/LiberoTutorial/Files
--- a/src/base/olo_base_dev.core
+++ b/src/base/olo_base_dev.core
@@ -50,6 +50,7 @@ filesets:
     files:
       - "tcl/olo_base_cc_simple.tcl" : {copyto: "base/olo_base_cc_simple.tcl"}
       - "tcl/olo_base_reset_gen.tcl" : {copyto: "base/olo_base_reset_gen.tcl"}
+      - "tcl/olo_base_ram_sdp.tcl" : {copyto: "base/olo_base_ram_sdp.tcl"}
       - "tcl/olo_base_cc_reset.tcl" : {copyto: "base/olo_base_cc_reset.tcl"}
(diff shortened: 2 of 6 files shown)


rbrglez:
> Topic 2
> I'd like to forward this to @rbrglez . Is there any problem in switchign from XDC to TCL? AFAIK they do the same - but I am not sure if this also applies for FuseSoC? Could you clarify?

I tested @tasgomes solution of adding `-unmanaged` to the TCL script when calling `read_xdc -ref` and it worked fine with FuseSoC.

Test Instructions:

1. Checkout the test branch

Clone and switch to the test branch:
```
git clone git@github.com:rbrglez/open-logic.git
git checkout test/scoped-constraints
```
Link to the branch on github:  [test/scoped-constraints](https://github.com/rbrglez/open-logic/tree/test/scoped-constraints)

2. Add FuseSoC libraries:

```bash
fusesoc library add src
fusesoc library add doc
```
3. Build the `vivado_tutorial` target

build `vivado_tutorial` with fusesoc:
```bash
fusesoc run --target zybo_z7 vivado_tutorial
```
4. Open generated vivado project

4.1. Check constraints:

image

4.2. Check vivado timing report.

 I check if `no_input_delay` has no warnings. This means that `olo_intf_sync` scoped constraint have been applied as expected. 

image

obruendl:
@tasgomes 

I still believe the constraint should apply to the SDP ram, not only to its usage in the FIFO. The same topic applies to all asynchronous DPRAM applications.

tasgomes:
Feature is completed and ready for review.

rbrglez:
Before you merge this feature to develop, fusesoc cores should be updated.
They can be updated by running following python script in `<oloRoot>/tools/fusesoc`:

```
python3 UpdateCoreFiles.py  --version 4.0.0 --cl-fix-version 2.2.1
```

PR #205, opened 2025-08-14
Add prescaling feature to I2C Master, including tests and documentation.

aruede:
This PR introduces a runtime-configurable prescaler for the I2C-Master, allowing dynamic adjustment of the SCL clock frequency via the Cmd_ClkPrescale port.
The feature is controlled by the PrescalerBits_g generic and maintains full backward compatibility when disabled (which is the default).

Documentation and interface descriptions have been updated accordingly.

### DIFF
--- a/test/tb/olo_test_i2c_vc.vhd
+++ b/test/tb/olo_test_i2c_vc.vhd
@@ -1027,7 +1027,7 @@ begin
                 else
                     expect_byte_excl_clock(data_slv, scl, sda,
                                            to_string(msg_p) & " - Receive byte in slave mode [I2cExpectRxByte]",
-                                           timeout => timeout,
+                                           timeout     => timeout,
                                            clk_stretch => clk_stretch);
                     send_bit_excl_clock(ack_output, scl, sda,
                                         to_string(msg_p) & " - ACK in slave mode [I2cExpectRxByte]",


doc/intf/olo_intf_i2c_master.md
```diff
@@ -57,6 +57,7 @@ tristate buffers.
 | :----------------- | :------ | ------- | :----------------------------------------------------------- |
 | ClkFrequency_g     | real    | -       | Frequency of the clock _Clk_ in Hz.<br />For correct operation, the clock frequency must be at least 16x higher than the I2C bus frequency. |
 | I2cFrequency_g     | real    | 100.0e3 | Frequency of the _I2cScl_ signal in Hz                       |
```
obruendl: I think this one deserves a comment that this is the SCL frequency without prescaling.

doc/intf/olo_intf_i2c_master.md
```diff
@@ -73,13 +74,14 @@ tristate buffers.
...
+| Cmd_Valid       | in     | 1               | -       | AXI-S handshaking signal for _Cmd_..._                       |
+| Cmd_Command     | in     | 3               | -       | Command to execute (_olo_intf_i2c_master_pkg_  constant names in brackets)<br />"000" =\> Send startcondition (_I2cCmd_Start_c_)<br />"001" =\> Send stop condition (_I2cCmd_Stop_c_)<br />"010" =\> Send rep. start condition"  (_I2cCmd_RepStart_c_)   <br />"011" =\> Send data  byte (_I2cCmd_Send_c_) <br />"100" =\> Receive data byte  (_I2cCmd_Receive_c_) |
+| Cmd_Data        | in     | 8               | -       | Data to send (only for _I2cCmd_Send_c_  resp. _Cmd_Command="011"_). |
+| Cmd_Ack         | in     | 1               |         | Acknowledge to send (only for _I2cCmd_Receive_c_ resp. _Cmd_Command="100")._<br />**'1'** =\> send ACK , **'0'** =\> send NACK. <br />**Note that polarity is inverted compared to the acknowledge bit on the I2C bus.** |
+| Cmd_ClkPrescale | in     | PrescalerBits_g | '0'     | Runtime prescaler value for SCL clock. If _PrescalerBits_g_ > 0, the SCL frequency _I2cFrequency_g_ is divided by 2^Cmd_ClkPrescale. No prescaling is applied when _Cmd_ClkPrescale_ = "0". |
```
obruendl: I thought quite a while about it and I concluded that we should divide not by "2^N" but by "N". The reason for this conclusion is that the traditional I2C frequencies are 100kHz, 400kHz and 1MHz. The factor between 1MHz and 100kHz is 10 (so not a power of two). I think we should define the prescaler in a way that one I2C master can handle at least all standard frequencies.

If we change to "divide by N", I think the naming "Prescale" does no more fit and "Cmd_SclDiv" would be better. Prescaling sounds like "2^N" (at least to me), which it currently is but it will no more be after the change.

src/intf/vhdl/olo_intf_i2c_master.vhd
```diff
@@ -103,8 +106,10 @@ architecture rtl of olo_intf_i2c_master is
 
     -- *** Constants ***
     constant BusyTimoutLimit_c    : integer := integer(ClkFrequency_g * BusBusyTimeout_g) - 1;
-    constant QuarterPeriodLimit_c : integer := integer(ceil(ClkFrequency_g / I2cFrequency_g / 4.0)) - 1;
+    constant QuarterPeriodLimit_c : integer := integer(ceil(ClkFrequency_g / I2cFrequency_g / 4.0));
```
obruendl: Why was the -1 removed?

src/intf/vhdl/olo_intf_i2c_master.vhd
```diff
@@ -103,8 +106,10 @@ architecture rtl of olo_intf_i2c_master is
...
     constant CmdTimeoutLimit_c    : integer := integer(ClkFrequency_g * CmdTimeout_g) - 1;
+    constant PrescalerMax_c       : integer := integer(real(2**(2**PrescalerBits_g - 1)));
```
obruendl: Why do you convert to _real_ and then back to _integer_?

src/intf/vhdl/olo_intf_i2c_master.vhd
```diff
@@ -179,7 +185,7 @@ begin
         v.QPeriodTick := '0';
         if (r.Fsm = BusIdle_s) or (r.Fsm = BusBusy_s) then
             v.QuartPeriodCnt := (others => '0');
-        elsif r.QuartPeriodCnt = QuarterPeriodLimit_c then
+        elsif r.QuartPeriodCnt = shift_left(to_unsigned(QuarterPeriodLimit_c, QuarterPeriodCntBits), to_integer(v.Prescaler)) - 1 then
```
obruendl: A barrel shifter on the right hand side of the comparison seems quite resource intense for what we do here. I'd rather suggest to have two nested counters (one for the basic SCL frequency and one for the division added in this PR).

With the change from _Prescale_ to _SclDiv_ where it is no more a power of two, the _shift_left_ approach won't work anymore anyways.

test/tb/olo_test_i2c_vc.vhd
```diff
@@ -563,7 +563,7 @@ architecture a of olo_test_i2c_vc is
         signal scl  : inout std_logic;
         signal sda  : inout std_logic;
         msg         : in    string;
-        timeout     : in    time := 1 ms;
+        timeout     : in    time := 10 ms;
```
obruendl: Why is this changed in the VC and not in the calling code? It's a default value, so it can be overridden where required.

test/intf/olo_intf_i2c_master/olo_intf_i2c_master_tb.vhd
obruendl, line 1: In the TB, could make the Prescale port dynamic ( so it only holds the correct value when a _I2cCmd_Start_c_ is injected and 'X' es otherwise)? Like this we test if the value actually only is sampled at the right point in time.

From the code I see this is the case - but covering it in the TB as well would be better for long-term maintenance resp. regression testing.

doc/intf/olo_intf_i2c_master.md
aruede: I agree, I have implemented the nested counter solution. I agree its the better solution in terms of resource utilization and also gives better control over the achieved frequency value.

obruendl:
@aruede 

Could you give me permission to push on the feature branch? I did fix the linter errors and one corner condition in VHDL. I'd like to apply those changes and then accept the PR.

Alternatively, you can merge my change from the _feature/i2c-master-prescaler_ branch on the open logic main repo to your branch (so they show up in this PR).

PR #212, opened 2025-09-01
WIP: Feature/hashtable

AnthoJack:
Adds a hashtable with support for read, write, remove, clear, and key list features

- [X] Documentation
- [x] Tests
- [x] Implementation

### DIFF
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -6,7 +6,11 @@
 
 ## Status Information
 
-VHDL Source: [olo_base_fifo_sync](../../src/base/vhdl/olo_base_hashtable.vhd)
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/coverage/olo_base_hashtable.json?cacheSeconds=0)
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/branches/olo_base_hashtable.json?cacheSeconds=0)
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/issues/olo_base_hashtable.json?cacheSeconds=0)
+
+VHDL Source: [olo_base_hashtable](../../src/base/vhdl/olo_base_hashtable.vhd)
 
 ## Description
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -25,10 +25,13 @@ The memory is described in a way that it utilizes RAM resources (Block-RAM or di
 | Depth_g           | positive  | -         | Number of storable elements. Must be a power of two |
 | KeyWidth_g        | positive  | -         | Width of key |
 | ValueWidth_g      | positive  | -         | Width of Value |
-|HashMode_g         | HashMode_t| DIVISION  | Hashing algorithm used |
+| Hash_g            | Hash_t    | DIVISION  | Hashing algorithm used |
+| RamStyle_g      | string   | "auto"  | Passed to [*olo_base_ram_sdp*](./olo_base_ram_sdp.md). Refer to the documentation of this component for more info |
+| RamBehavior_g   | string   | "RBW"   | Passed to [*olo_base_ram_sdp*](./olo_base_ram_sdp.md). Refer to the documentation of this component for more info  |
 
-The width of the internal RAM is calculated using the following equation: RamWidth = KeyWidth_g + ValueWidth_g + 1
-The supplementary bit comes from the "used" bit that memorises which words contain valid key-value pairs
+Current supported hash algorithms are:
+* *DIVISION*: Key's value is used as-is, except for modulo against depth of hashtable to obtain a valid index
+* *CRC*: Key's value is hashed using *olo_base_crc*
 
 ## Interfaces
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -73,7 +73,7 @@ Current supported hash algorithms are:
 | Out_Ready         | out    | 1                | N/A     | Ready for new operation                                         |
 | Out_KeyUnknown    | out    | 1                | N/A     | _In_Key_ does not exist in the hashtable (WR/RM/RD operations)  |
 | Out_Full          | out    | 1                | N/A     | Full                                                            |
-| Out_Size          | out    | _IndexWidth_g_+1 | N/A     | Number of Key-Value pairs stored |
+| Out_Pairs          | out    | _IndexWidth_g_+1 | N/A     | Number of Key-Value pairs stored |
 
 
 ## Architecture
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -88,7 +88,9 @@ Current supported hash algorithms are:
 
 #### Key search
 
-The hastable uses a hash function to obtain a memory index from the provided key. As the hashing could produce the same result for different keys,the hashtable uses [linear open-addressing](https://en.wikipedia.org/wiki/Open_addressing) to resolve those collisions: the following elements are checked until either the key or an empty spot is found, creating clusters (or search-chains). When a key is found, _Out_KeyUnknown_ is set to '0' and to '1' otherwise. A counter is also used to prevent endless looping when memory is full as there would be no empty spot to end the search on
+The hastable uses a hash function to obtain a memory index from the provided key. As the hashing could produce the same result for different keys,the hashtable uses [linear open-addressing](https://en.wikipedia.org/wiki/Open_addressing) to resolve those collisions: the following elements are checked until either the key or an empty spot is found, creating clusters (or search-chains). When a key is found, _Out_KeyUnknown_ is reset to '0' and set to '1' otherwise. A counter is also used to prevent endless looping when memory is full as there would be no empty spot to end the search on
+
+The more pairs are stored, the more chances of collisions and thus the formation of clusters whose size rapidly reduces the performance of the hashtable. When performance is an important factor, the hashtable should be dimensioned such that it never comes close to being full 
 
 #### Read
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -45,50 +45,40 @@ Current supported hash algorithms are:
 | Clk  | in     | 1      | -       | Clock                                           |
 | Rst  | in     | 1      | -       | Reset input (high-active, synchronous to _Clk_) |
 
-### Input Data
+### Input
 
-| Name     | In/Out | Length        | Default   | Description                                  |
-| :------- | :----- | :--------     | -------   | :------------------------------------------- |
-| In_Key   | in     | _KeyWidth_g_   | -         | Input key                                    |
-| In_Value | in     | _ValueWidth_g_ | -         | Input Value                                  |
-
-### Output Data
-
-| Name      | In/Out | Length           | Default | Description                                   |
-| :-------- | :----- | :--------        | ------- | :-------------------------------------------- |
-| Out_Key   | out    | _KeyWidth_g_      | N/A     | Stored key output |
-| Out_Value | out    | _ValueWidth_g_    | N/A     | Output value                                  |
-
-
-### Operations
-
-| Name          | In/Out | Length    | Default   | Latency (ticks)| Description |
-| :-------      | :----- | :-------- | :-------   | :---- | :-------------------------------------------    |
+| Name     | In/Out | Length        | Default   | Latency (ticks) | Description                                  |
+| :------- | :----- | :--------     | -------   | :--- | :------------------------------------------- |
+| In_Key   | in     | _KeyWidth_g_   | -         | N/A | Input key                                    |
+| In_Value | in     | _ValueWidth_g_ | -         | N/A | Input Value                                  |
 | In_Write      | in     | 1         | -         | search_key + 1 | (Over)Write In_Key-In_Value pair                 |
 | In_Read       | in     | 1         | -         | search_key + Out_Ready_delay + 1 | Read Out_Value corresponding to In_Key           |
 | In_Remove     | in     | 1         | -         | search_key + cluster_comp + 1 | Remove Key-Value pair corresponding to In_Key    |
 | In_Clear      | in     | 1         | -         | Width_g | Clear All Key-Value pairs from memory            |
 | In_NextKey    | in     | 1         | -         | distance_from_next_key | Find next valid Out_Key in memory                |
+| In_Ready      | out    | 1         | -         | N/A. | Axi-Stream handshaking signal that hashtable is ready to accept an operation |
+| In_Valid      | in     | 1         | -         | N/A. | Axi_Stream handshaking signal that an operation is requested |
 
 Operations are given in order of priority: if multiple operations are requested, only the highest in the list is performed
 
 Latency of search_key is: *1 + element_position_in_cluster*
+
 Latency of *cluster_comp* is: *1 + (cluster_size - element_position_in_cluster)*
 
-### Handshaking
+### Output
 
-| Name | In/Out | Length | Default | Description |
-| :--- | :--- | :--- | :--- | :--- |
-| In_Ready      | out    | 1         | -         | Axi-Stream handshaking signal that hashtable is ready to accept an operation |
-| In_Valid      | in     | 1         | -         | Axi_Stream handshaking signal that an operation is requested |
+| Name      | In/Out | Length           | Default | Description                                   |
+| :-------- | :----- | :--------        | ------- | :-------------------------------------------- |
+| Out_Key   | out    | _KeyWidth_g_      | N/A     | Stored key output |
+| Out_Value | out    | _ValueWidth_g_    | N/A     | Output value                                  |
 | Out_Ready      | in    | 1         | -         | Axi-Stream handshaking signal that user is ready to accept data |
 | Out_Valid      | out     | 1         | -         | Axi_Stream handshaking signal that hashtable has data for the user |
+| Out_KeyUnknown    | out    | 1                | N/A     | _In_Key_ does not exist in the hashtable (WR/RM/RD operations)  |
 
 ### Status
 
 | Name              | In/Out | Length           | Default | Description                                                     |
 | :--------         | :----- | :--------------- | ------- | :----------------------                                         |
-| Out_KeyUnknown    | out    | 1                | N/A     | _In_Key_ does not exist in the hashtable (WR/RM/RD operations)  |
 | Status_Full          | out    | 1                | N/A     | Full                                                            |
 | Status_Pairs          | out    | _log2ceil(Depth_g)_+1 | N/A     | Number of Key-Value pairs stored |
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -85,6 +85,7 @@ Latency of *cluster_comp* is: *1 + (cluster_size - element_position_in_cluster)*
 
 | Name              | In/Out | Length           | Default | Description                                                     |
 | :--------         | :----- | :--------------- | ------- | :----------------------                                         |
+| Status_Busy       | out | 1 | N/A | Busy |
 | Status_Full          | out    | 1                | N/A     | Full                                                            |
 | Status_Pairs          | out    | _log2ceil(Depth_g)_+1 | N/A     | Number of Key-Value pairs stored |
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -165,7 +165,26 @@ Note also that memory is used cyclically and thus, elements at indices 7 and 0 w
 
 wr_idx < rd_idx && (hash <= wr_idx || hash > rd_idx) || rd_idx <= wr_idx && hash > rd_idx && hash <= wr_idx
 
-Where `wr_idx` is the index of the current "hole" in the cluster, `rd_idx` is a following element of the cluster and `hash` is the hash of the element at `rd_idx`. When the equation is true, the element at `rd_idx` is copied at `wr_idx`. Once an empty element is found, the cluster is complete and the last copied element is emptied
+Where `wr_idx` is the index of the current "hole" in the cluster, `rd_idx` is a following element of the cluster and `hash` is the hash of the element at `rd_idx`. When the equation is true, the element at `rd_idx` is copied at `wr_idx`. Once an empty element is found, the cluster is complete and the last copied element is emptied. Here's a more complete example using compression-first as explained above:
+
+Idx | Init  | Compression | Removal | Cmp | Rm | Cmp | Rm | Cmp | Rm | Cmp | Rm
+---: | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :---
+0   |       |       |       |       |       |       |       |       |       |       | 
+1   |       |       |       |       |       |       |       |       |       |       | 
+2   |       |       |       |       |       |       |       |       |       |       | 
+3   |       |       |       |       |       |       |       |       |       |       | 
+4   | A(4)  | A(4)  | *A(4)*  | C(4)  | *C(4)*  | E(4)  | *E(4)*  | J(4)  | ***J(4)***  |    
+5   | *B(4)*  | C(4)  | C(4)  | D(5)  | D(5)  | D(5)  | D(5)  | D(5)  | D(5)  | D(5)  | D(5)  
+6   | C(4)  | D(5)  | D(5)  | E(4)  | E(4)  | F(5)  | F(5)  | F(5)  | F(5)  | F(5)  | F(5)  
+7   | D(5)  | E(4)  | E(4)  | F(5)  | F(5)  | G(7)  | G(7)  | G(7)  | G(7)  | G(7)  | G(7)  
+8   | E(4)  | F(5)  | F(5)  | G(7)  | G(7)  | H(5)  | H(5)  | H(5)  | H(5)  | H(5)  | H(5)  
+9   | F(5)  | G(7)  | G(7)  | H(5)  | H(5)  | J(4)  | J(4)  | **J(4)**  |       |       | 
+10  | G(7)  | H(5)  | H(5)  | I(10) | I(10) | I(10) | I(10) | I(10) | I(10) | I(10) | I(10)
+11  | H(5)  | I(10) | I(10) | J(4)  | J(4)  | **J(4)**  |       |       |       |       | 
+12  | I(10) | J(4)  | J(4)  | **J(4)**  |       |       |       |       |       |       | 
+13  | J(4)  | **J(4)**  |       |       |       |       |       |       |       |       | 
+14  |       |       |       |       |       |       |       |       |       |       | 
+15  |       |       |       |       |       |       |       |       |       |       | 
 
 In the special case where an element must be removed from a full hashtable, the state machine would run endlessly as there is no empty memory spot to signal the end of a cluster. To prevent this, a counter is used to detect when the memory was fully checked
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -14,10 +14,9 @@ VHDL Source: [olo_base_hashtable](../../src/base/vhdl/olo_base_hashtable.vhd)
 
 ## Description
 
-This component implements a synchronous [hashtable](https://en.wikipedia.org/wiki/Hash_table) capable of storing data using an integer value as key.
+This component implements a synchronous [hashtable](https://en.wikipedia.org/wiki/Hash_table) capable of storing *std_logic_vector* key-value pairs.
 
-The memory is described in a way that it utilizes RAM resources (Block-RAM or distributed RAM) available in FPGAs with
-commonly used tools. For this purpose [olo_base_ram_sdp](./olo_base_ram_sdp.md) is used.
+The memory is described in a way that it utilizes RAM resources (Block-RAM or distributed RAM) available in FPGAs with commonly used tools. For this purpose [olo_base_ram_sdp](./olo_base_ram_sdp.md) is used.
 
 ## Generics
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -58,13 +58,13 @@ Current supported hash algorithms are:
 
 ### Operations
 
-| Name          | In/Out | Length    | Default   | Description                                      |
-| :-------      | :----- | :-------- | -------   | :-------------------------------------------     |
-| In_Write      | in     | 1         | -         | (Over)Write In_Key-In_Value pair                 |
-| In_Read       | in     | 1         | -         | Read Out_Value corresponding to In_Key           |
-| In_Remove     | in     | 1         | -         | Remove Key-Value pair corresponding to In_Key    |
-| In_Clear      | in     | 1         | -         | Clear All Key-Value pairs from memory            |
-| In_NextKey    | in     | 1         | -         | Find next valid Out_Key in memory                |
+| Name          | In/Out | Length    | Default   | Latency | Description |
+| :-------      | :----- | :-------- | -------   | :---- | :-------------------------------------------    |
+| In_Write      | in     | 1         | -         | TODO | (Over)Write In_Key-In_Value pair                 |
+| In_Read       | in     | 1         | -         | TODO | Read Out_Value corresponding to In_Key           |
+| In_Remove     | in     | 1         | -         | TODO | Remove Key-Value pair corresponding to In_Key    |
+| In_Clear      | in     | 1         | -         | TODO | Clear All Key-Value pairs from memory            |
+| In_NextKey    | in     | 1         | -         | TODO | Find next valid Out_Key in memory                |
 
 ### Status
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -1,0 +1,160 @@
+<img src="../Logo.png" alt="Logo" width="400">
+
+# olo_base_hashtable
+
+[Back to **Entity List**](../EntityList.md)
+
+## Status Information
+
+VHDL Source: [olo_base_fifo_sync](../../src/base/vhdl/olo_base_hashtable.vhd)
+
+## Description
+
+This component implements a synchronous [hashtable](https://en.wikipedia.org/wiki/Hash_table) capable of storing data using an integer value as key.
+
+The memory is described in a way that it utilizes RAM resources (Block-RAM or distributed RAM) available in FPGAs with
+commonly used tools. For this purpose [olo_base_ram_sdp](./olo_base_ram_sdp.md) is used.
+
+## Generics
+
+| Name              | Type      | Default   | Description                                                  |
+| :--------------   | :-------- | -------   | :----------------------------------------------------------- |
+| IdxWidth_g        | positive  | -         | Width of indices. Depth of internal RAM is set using this parameter (RamDepth = 2**IdxWidth_g) |
+| KeyWidth_g        | positive  | -         | Width of key |
+| ValueWidth_g      | positive  | -         | Width of Value |
+|HashMode_g         | HashMode_t| DIVISION  | Hashing algorithm used |
+
+The width of the internal RAM is calculated using the following equation: RamWidth = KeyWidth_g + ValueWidth_g + 1
+The supplementary bit comes from the "used" bit that memorises which words contain valid key-value pairs
+
+## Interfaces
+
+### Synchronisation
+
+| Name | In/Out | Length | Default | Description                                     |
+| :--- | :----- | :----- | ------- | :---------------------------------------------- |
+| Clk  | in     | 1      | -       | Clock                                           |
+| Rst  | in     | 1      | -       | Reset input (high-active, synchronous to _Clk_) |
+
+### Input Data
+
+| Name     | In/Out | Length        | Default   | Description                                  |
+| :------- | :----- | :--------     | -------   | :------------------------------------------- |
+| In_Key   | in     | _KeyWidth_g_   | -         | Input key                                    |
+| In_Value | in     | _ValueWidth_g_ | -         | Input Value                                  |
+
+### Output Data
+
+| Name      | In/Out | Length           | Default | Description                                   |
+| :-------- | :----- | :--------        | ------- | :-------------------------------------------- |
+| Out_Value | out    | _ValueWidth_g_    | N/A     | Output value                                  |
+| Out_Key   | out    | _KeyWidth_g_      | N/A     | Stored key output                             |
+
+### Operations
+
+| Name          | In/Out | Length    | Default   | Description                                      |
+| :-------      | :----- | :-------- | -------   | :-------------------------------------------     |
+| In_Write      | in     | 1         | -         | (Over)Write In_Key-In_Value pair                 |
+| In_Read       | in     | 1         | -         | Read Out_Value corresponding to In_Key           |
+| In_Remove     | in     | 1         | -         | Remove Key-Value pair corresponding to In_Key    |
+| In_Clear      | in     | 1         | -         | Clear All Key-Value pairs from memory            |
+| In_NextKey    | in     | 1         | -         | Find next valid Out_Key in memory                |
+
+### Status
+
+| Name              | In/Out | Length           | Default | Description                                                     |
+| :--------         | :----- | :--------------- | ------- | :----------------------                                         |
+| Out_Ready         | out    | 1                | N/A     | Ready for new operation                                         |
+| Out_KeyUnknown    | out    | 1                | N/A     | _In_Key_ does not exist in the hashtable (WR/RM/RD operations)  |
+| Out_Full          | out    | 1                | N/A     | Full                                                            |
+| Out_Size          | out    | _IndexWidth_g_+1 | N/A     | Number of Key-Value pairs stored |
+
+
+## Architecture
+
+**Note**: Operations requested while hashtable isn't ready (_Out_Ready_ = '1') are ignored
+
+### Key search
+
+The hastable uses a hash function to obtain a memory index from the provided key. As the hashing could produce the same result for different keys,the hashtable uses [linear open-addressing](https://en.wikipedia.org/wiki/Open_addressing) to resolve those collisions: the following elements are checked until either the key or an empty spot is found, creating clusters (or search-chains). When a key is found, _Out_KeyUnknown_ is set to '0' and to '1' otherwise. A counter is also used to prevent endless looping when memory is full as there would be no empty spot to end the search on
+
+### Read
+
+Read searches for _In_Key_ in the hashtable and updates _Out_Value_ if the key is found
+
+### Write 
+
+Write searches for _In_Key_ (or an empty spot if the element isn't already in the hashtable) and then writes the _In_Value_ and _In_Key_ in memory along with a `used` value to signal that this spot is occupied
+
+When hashtable is full (_Out_Full_ = '1'), it is still possible to overwrite existing keys. In such cases, use _Out_KeyUnknown_ to known if new value was overwritten (_Out_KeyUnknown_ = '0') or ignored (_Out_KeyUnknown_ = '1')
+
+### Removal 
+
+Removal searches for _In_Key_ and (if key is found) sets the `used` bit to 0 before performing cluster compression (see below)
+
+### Clear
+
+Clear goes through the whole memory and clears all `used` values
+
+### Next Key
+
+Next Key sequentially and cyclically goes through the hash table looking for the next used spot and updates _Out_Key_ with the value. Modifying the internal memory may disturb the recovery of all the keys. To reliably obtain all values stored in the hashtable, use the following steps:
+
+1. Invoke In_NextKey
+2. Wait till hashtable is ready again to read Out_key
+3. Repeat steps 1 and 2 _Out_Size_ times
+
+### Cluster compression
+
+When removing key-value pairs, holes may be created in those clusters, rendering some elements inaccessible. Take for example the following situation:
+
+Index | Data
+---: | :---
+0 | -
+1 | -
+2 | Data_1, hash = 2
+3 | Data_2, hash = 2
+4 | Data_3, hash = 3
+5 | Data_4, hash = 5
+6 | -
+7 | -
+
+When searching for Data_3, hash would indicate to go search at index 3 but since Data_2 is already there, hashtable looks at the following index and finds Data_3. Trying to remove the element at index 3 would prevent further searches for Data_3 from succeeding
+
+Index | Data
+---: | :---
+0 | -
+1 | -
+2 | Data_1, hash = 2
+3 | -
+4 | Data_3, hash = 3 (now inaccessible)
+5 | Data_4, hash = 5
+6 | -
+7 | -
+
+When searching for Data_3, hash would redirect to index 3 but since memory is empty here, the hashtable would believe that Data_3 doesn't exist. Search chain must then be rebuilt by copying subsequent elements back into the cluster if their hash corresponds
+
+Index | Data
+---: | :---
+0 | -
+1 | -
+2 | Data_1, hash = 2
+3 | Data_3, hash = 3 (accessible again)
+4 | -
+5 | Data_4, hash = 5 (not moved because already at the right place)
+6 | -
+7 | -
+
+Note that in this implementation, only the last element of the cluster is actually removed (memory written to 0). In practice, the cluster is first rebuilt, (possibly squashing the data to remove) and then the last element of the cluster (possibly the element to remove if it was the sole element of the cluster) is emptied. This strategy simplifies the state machine necessary to remove an element
+
+Note also that memory is used cyclically and thus, elements at indices 7 and 0 would be considered as part of the same cluster. When reconstituting the cluster, the following equation is used to determine if an element's hash places it outside the search chain (cluster broken by remove):
+
+$
+wr\_idx < rd\_idx \wedge (hash <= wr\_idx \vee hash > rd\_idx) \vee rd\_idx <= wr\_idx \wedge hash > rd\_idx \wedge hash <= wr\_idx
+$
+
+Where `wr_idx` is the index of the current "hole" in the cluster, `rd_idx` is a following element of the cluster and `hash` is the hash of the element at `rd_idx`. When the equation is true, the element at `rd_idx` is copied at `wr_idx`. Once an empty element is found, the cluster is complete and the last copied element is emptied
+
+In the special case where an element must be removed from a full hashtable, the state machine would run endlessly as there is no empty memory spot to signal the end of a cluster. To prevent this, a counter is used to detect when the memory was fully checked
+
+
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -90,25 +90,25 @@ Current supported hash algorithms are:
 
 The hastable uses a hash function to obtain a memory index from the provided key. As the hashing could produce the same result for different keys,the hashtable uses [linear open-addressing](https://en.wikipedia.org/wiki/Open_addressing) to resolve those collisions: the following elements are checked until either the key or an empty spot is found, creating clusters (or search-chains). When a key is found, _Out_KeyUnknown_ is set to '0' and to '1' otherwise. A counter is also used to prevent endless looping when memory is full as there would be no empty spot to end the search on
 
-### Read
+#### Read
 
 Read searches for _In_Key_ in the hashtable and updates _Out_Value_ if the key is found
 
-### Write 
+#### Write 
 
 Write searches for _In_Key_ (or an empty spot if the element isn't already in the hashtable) and then writes the _In_Value_ and _In_Key_ in memory along with a `used` value to signal that this spot is occupied
 
 When hashtable is full (_Out_Full_ = '1'), it is still possible to overwrite existing keys. In such cases, use _Out_KeyUnknown_ to known if new value was overwritten (_Out_KeyUnknown_ = '0') or ignored (_Out_KeyUnknown_ = '1')
 
-### Removal 
+#### Removal 
 
 Removal searches for _In_Key_ and (if key is found) sets the `used` bit to 0 before performing cluster compression (see below)
 
-### Clear
+#### Clear
 
 Clear goes through the whole memory and clears all `used` values
 
-### Next Key
+#### Next Key
 
 Next Key sequentially and cyclically goes through the hash table looking for the next used spot and updates _Out_Key_ with the value. Modifying the internal memory may disturb the recovery of all the keys. To reliably obtain all values stored in the hashtable, use the following steps:
 
@@ -116,7 +116,7 @@ Next Key sequentially and cyclically goes through the hash table looking for the
 2. Wait till hashtable is ready again to read Out_key
 3. Repeat steps 1 and 2 _Out_Size_ times
 
-### Cluster compression
+## Cluster compression
 
 When removing key-value pairs, holes may be created in those clusters, rendering some elements inaccessible. Take for example the following situation:
 
--- a/doc/base/olo_base_hashtable.md
+++ b/doc/base/olo_base_hashtable.md
@@ -28,6 +28,7 @@ The memory is described in a way that it utilizes RAM resources (Block-RAM or di
 | Hash_g            | Hash_t    | DIVISION  | Hashing algorithm used |
 | RamStyle_g      | string   | "auto"  | Passed to [*olo_base_ram_sdp*](./olo_base_ram_sdp.md). Refer to the documentation of this component for more info |
 | RamBehavior_g   | string   | "RBW"   | Passed to [*olo_base_ram_sdp*](./olo_base_ram_sdp.md). Refer to the documentation of this component for more info  |
+| ClearAfterReset_g | boolean | true | Clear memory after a reset to prevent erroneous values |
 
 Current supported hash algorithms are:
 * *DIVISION*: Key's value is used as-is, except for modulo against depth of hashtable to obtain a valid index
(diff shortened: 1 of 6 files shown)


doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
+<img src="../Logo.png" alt="Logo" width="400">
+
+# olo_base_hashtable
+
+[Back to **Entity List**](../EntityList.md)
+
+## Status Information
+
+VHDL Source: [olo_base_fifo_sync](../../src/base/vhdl/olo_base_hashtable.vhd)
```
obruendl: Please add the badges links (copy from any other documenation and adjust the name).

You do not have to create the badges. They will be created as part of the coverage simulation when you code is mainlined.

obruendl: The text still says _olo_base_fifo_sync_

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Description
+
+This component implements a synchronous [hashtable](https://en.wikipedia.org/wiki/Hash_table) capable of storing data using an integer value as key.
```
obruendl: It's not an "integer". Suggested formulation "... storing std_logic_vector key/value pairs."

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+The memory is described in a way that it utilizes RAM resources (Block-RAM or distributed RAM) available in FPGAs with
+commonly used tools. For this purpose [olo_base_ram_sdp](./olo_base_ram_sdp.md) is used.
+
+## Generics
+
+| Name              | Type      | Default   | Description                                                  |
+| :--------------   | :-------- | -------   | :----------------------------------------------------------- |
+| IdxWidth_g        | positive  | -         | Width of indices. Depth of internal RAM is set using this parameter (RamDepth = 2**IdxWidth_g) |
```
obruendl: Open Logic everywhere used "Depth_g" (so the number of entries in a storage, _not_ the number of address bits). For consistency I suggest you do so here ase well.

I fully get that the implementation requires a power-of-two because the hash might be any value with a certain width. This is okay - you can still document that depth must be a power of two and check if this is the case using an assertion inside the entity.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+| KeyWidth_g        | positive  | -         | Width of key |
+| ValueWidth_g      | positive  | -         | Width of Value |
+|HashMode_g         | HashMode_t| DIVISION  | Hashing algorithm used |
```
obruendl: The different options should be documented somewhere

obruendl: I am not sure if _mode_ is a good wording here. Probably _HashAlgo_g_ or just _Hash_g_ would be a better fit. However, if you prefer _HashMode_g_ this is also fine. I just wanted to raise the point - I don't think it affects usability much.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+The width of the internal RAM is calculated using the following equation: RamWidth = KeyWidth_g + ValueWidth_g + 1
```
obruendl: This belongs into architecture. It is purely internal IMO.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+VHDL Source: [olo_base_fifo_sync](../../src/base/vhdl/olo_base_hashtable.vhd)
+
+## Description
+
+This component implements a synchronous [hashtable](https://en.wikipedia.org/wiki/Hash_table) capable of storing data using an integer value as key.
+
+The memory is described in a way that it utilizes RAM resources (Block-RAM or distributed RAM) available in FPGAs with
+commonly used tools. For this purpose [olo_base_ram_sdp](./olo_base_ram_sdp.md) is used.
+
+## Generics
```
obruendl, line 21: Please make _RamStyle_g_ and _RamBehavior_g_ from the SDP RAM available in your entity as well. These generics are required to control RAM inference for different technologies.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+## Interfaces
+
+### Synchronisation
+
+| Name | In/Out | Length | Default | Description                                     |
+| :--- | :----- | :----- | ------- | :---------------------------------------------- |
+| Clk  | in     | 1      | -       | Clock                                           |
+| Rst  | in     | 1      | -       | Reset input (high-active, synchronous to _Clk_) |
+
+### Input Data
```
obruendl: Before writing the actual comment: That we discuss the interface up-front is they key idea behind doing doc first. So I expect that we have a little discussion here. I am not always right with every of my comments and I don't expect you to just silently follow every suggestion - I more appreciate a productive discussion with the best possible outcome (i.e. the most logical itnerface for users).

The HashTable has a single FSM and can only execute one operation at a time IIRC.

I'd therefore suggest to group the interface as follows. I also list which of the currently mentioned ports should go into which group.

* Request_Xxxx - Requesting an operation
  * In_Write / In_Read / In_Remove / In_Clear / In_NextKey (these 
  * In_Key / In_Vaue 
  * New: Valid / Ready handshaking signals (AXI4-Stream style)
* Out_Xxxx - Read results
  * Out_Value / Out_Key
  * Out_KeyUnknown
  * New: Valid handshaking signal (AXI4-Stream style)
  * Optional: Ready (AXI4-Stream style handshaking).
    * If you can easily check this signal in your FSM and wait until the output was accepted this would be great. However, iif this adds lots of complexity it also is OK to omit _Ready_ and document that the user code must accept the data immediately.
* Status_Xxxx - Asynchronous status information
  * Out_Full
  * Out_Size - I'd suggest to rename this from _Size_ to _Pairs_ or _Level_

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+# olo_base_hashtable
+
+[Back to **Entity List**](../EntityList.md)
+
+## Status Information
+
+VHDL Source: [olo_base_fifo_sync](../../src/base/vhdl/olo_base_hashtable.vhd)
+
+## Description
```
obruendl, line 15: Can you add the latency for every operation? Most operations depend on search cycles - so I'd suggest to indicate the minimum number of clock cycles required plust a statement if this is constant or variable.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Note**: Operations requested while hashtable isn't ready (_Out_Ready_ = '1') are ignored
+
+### Key search
+
+The hastable uses a hash function to obtain a memory index from the provided key. As the hashing could produce the same result for different keys,the hashtable uses [linear open-addressing](https://en.wikipedia.org/wiki/Open_addressing) to resolve those collisions: the following elements are checked until either the key or an empty spot is found, creating clusters (or search-chains). When a key is found, _Out_KeyUnknown_ is set to '0' and to '1' otherwise. A counter is also used to prevent endless looping when memory is full as there would be no empty spot to end the search on
+
+### Read
+
+Read searches for _In_Key_ in the hashtable and updates _Out_Value_ if the key is found
```
obruendl: I suggest mentioning the meaning or _KeyUnknown_ for every command. For _Read_ it means _Value_ is invalid because the key did not exist, for _Write_ it means "a new entry was created", etc.

I also suggest grouping all the commands in a separate sub-sections "operation" or the like. So we have the command descriptions separated from the general architecture.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+| Out_Size          | out    | _IndexWidth_g_+1 | N/A     | Number of Key-Value pairs stored |
+
+
+## Architecture
+
+**Note**: Operations requested while hashtable isn't ready (_Out_Ready_ = '1') are ignored
+
+### Key search
+
+The hastable uses a hash function to obtain a memory index from the provided key. As the hashing could produce the same result for different keys,the hashtable uses [linear open-addressing](https://en.wikipedia.org/wiki/Open_addressing) to resolve those collisions: the following elements are checked until either the key or an empty spot is found, creating clusters (or search-chains). When a key is found, _Out_KeyUnknown_ is set to '0' and to '1' otherwise. A counter is also used to prevent endless looping when memory is full as there would be no empty spot to end the search on
```
obruendl: Although being clear to most users I would add a note that due to the search chains performance depends on the number of entries and for good performance the table should never get anywhere close to full.

Intention here is to make this clear to less experienced users that just want some kind of table.

A few lines more documentation is better than answering many issues.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Read
+
+Read searches for _In_Key_ in the hashtable and updates _Out_Value_ if the key is found
+
+### Write 
+
+Write searches for _In_Key_ (or an empty spot if the element isn't already in the hashtable) and then writes the _In_Value_ and _In_Key_ in memory along with a `used` value to signal that this spot is occupied
```
obruendl: _used_ is mentioned first here. I think somehwere in the architecture it should be described.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+When hashtable is full (_Out_Full_ = '1'), it is still possible to overwrite existing keys. In such cases, use _Out_KeyUnknown_ to known if new value was overwritten (_Out_KeyUnknown_ = '0') or ignored (_Out_KeyUnknown_ = '1')
```
obruendl: With _Full_ becoming a asynchronous status signal (see comment in the ports section), we should add some indication in the _Out_ group to indicate if the value was actually written. IMO it would be best to add _Out_Error_ and document for every command if this is used and what it means - I am not sure if there are other cases where it is required, bot for writing into the full hash-table if would be used for sure.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Removal 
+
+Removal searches for _In_Key_ and (if key is found) sets the `used` bit to 0 before performing cluster compression (see below)
+
+### Clear
+
+Clear goes through the whole memory and clears all `used` values
```
obruendl: I suggest to add a statement from the users perspective that says "Clear does clear the whole table (i.e. it removes all key/value pairs)".

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Next Key
+
+Next Key sequentially and cyclically goes through the hash table looking for the next used spot and updates _Out_Key_ with the value. Modifying the internal memory may disturb the recovery of all the keys. To reliably obtain all values stored in the hashtable, use the following steps:
```
obruendl, line 126: I guess this does not only assign _Out_Key_ but also _Out_Value_ right? This should be mentioned.

How does the user know when the end of the table is reached? Yes, he can do so by looking for the first key to show up again - however, I signal could help.

We might discuss if we want to use the _Out_Error_ proposed further up - you could assign it in the first output after the wraparound.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+Index | Data
+---: | :---
+0 | -
+1 | -
+2 | Data_1, hash = 2
+3 | Data_2, hash = 2
+4 | Data_3, hash = 3
+5 | Data_4, hash = 5
+6 | -
```
obruendl: For clarity, can you add "Data_5" with "hash = 6" to index 6? 

This would make more clear that not just the "last element" is moved but the last one not braking the chain.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+0 | -
+1 | -
+2 | Data_1, hash = 2
+3 | Data_3, hash = 3 (accessible again)
+4 | -
+5 | Data_4, hash = 5 (not moved because already at the right place)
+6 | -
+7 | -
+
+Note that in this implementation, only the last element of the cluster is actually removed (memory written to 0). In practice, the cluster is first rebuilt, (possibly squashing the data to remove) and then the last element of the cluster (possibly the element to remove if it was the sole element of the cluster) is emptied. This strategy simplifies the state machine necessary to remove an element
```
obruendl, line 175: I have to think through the cluster compression in detail. I'll do this in the next round because I rund out of time for the review today.

obruendl (changes requested):

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Write 
+
+Write searches for _In_Key_ (or an empty spot if the element isn't already in the hashtable) and then writes the _In_Value_ and _In_Key_ in memory along with a `used` value to signal that this spot is occupied
+
+When hashtable is full (_Out_Full_ = '1'), it is still possible to overwrite existing keys. In such cases, use _Out_KeyUnknown_ to known if new value was overwritten (_Out_KeyUnknown_ = '0') or ignored (_Out_KeyUnknown_ = '1')
+
+### Removal 
+
+Removal searches for _In_Key_ and (if key is found) sets the `used` bit to 0 before performing cluster compression (see below)
```
obruendl: There are two options for cluster compression that seem sane to me:

1. Always compress clusters when removing an element. This is most handy for the user because he does not have to care about it. However, with high utilization it also can make removals taking longer.

2. Use _thombstones_ for gaps in clusters instead of always clean up immediately and let the user decide when to compress clusters (and then walk through the whole table, compressing all clusters). This would give the user better control.

Currently the plan is only to do option 1. Do we stay with this or would it be straightforward to implement both options?

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,160 @@
...
+5 | Data_4, hash = 5 (not moved because already at the right place)
+6 | -
+7 | -
+
+Note that in this implementation, only the last element of the cluster is actually removed (memory written to 0). In practice, the cluster is first rebuilt, (possibly squashing the data to remove) and then the last element of the cluster (possibly the element to remove if it was the sole element of the cluster) is emptied. This strategy simplifies the state machine necessary to remove an element
+
+Note also that memory is used cyclically and thus, elements at indices 7 and 0 would be considered as part of the same cluster. When reconstituting the cluster, the following equation is used to determine if an element's hash places it outside the search chain (cluster broken by remove):
+
+$
+wr\_idx < rd\_idx \wedge (hash <= wr\_idx \vee hash > rd\_idx) \vee rd\_idx <= wr\_idx \wedge hash > rd\_idx \wedge hash <= wr\_idx
```
obruendl: The formula is not rendered correctly. However, this is a minor thing. For the comment I just use && / ||.

I think the formula requires a bit of comment.

_wr_idx < ird_idx && (hash <= wr_idx || hash > rd_idx) || **rd_idx <= wr_idx && hash > rd_idx and hash <= wr_idx**_

To my understanding the bold part above covers search chains that wrap around from the highest address to zero and the non-bold part the common case. Correct?

For the common case part, I have a questions:

_wr_idx < rd_idx && (hash <= wr_idx || **hash > rd_idx**)_ 
What is the bold part for? To my understanding this should never happen because elements are written to the hash address or a higher address (if the address is copuied). The wraparound is handled by the other half of the expression anyways.

I did a little example, see below. Could you check if my understanding is correct. I'd suggest to add something like this example in the documentation - I think it helps the reader to understand what happens step by step.

image

obruendl (changes requested):
@AnthoJack 

I have completed the doc review. You can follow up from here.

doc/base/olo_base_hashtable.md
AnthoJack: I definitely can add the column now but filling in the values will have to wait until the very end I think since I can't really give a number until the hashtable is done and tested with the new interface/internal implementation. The values are more of a result rather than an constraint for the implementation so they'll have to be filled in later

doc/base/olo_base_hashtable.md
AnthoJack: *KeyUnknown* is only relevant for operations that require a search so I can definitely mention it for rd/wr/rm but it seems irrelevant for the others. Or should I explicitely state that it is irrelevant for those operations ?

doc/base/olo_base_hashtable.md
AnthoJack: I described it in the "Notes" at the beginning of the architecture. I also added another note indicating that memory should be cleared after reset (if not already guaranteed by some other mecanism) to prevent it from being set to 1 when hashtable is supposed to be empty

doc/base/olo_base_hashtable.md
AnthoJack: I'm not sure what you mean by "asynchronous" here since full and size will both always depend on the number of pairs stored which is tracked using a counter (synchronous by design), we'll discuss it in the related comment. When it comes to a _Out_Err_ signal. I don't know if it is important: No error may arise which would be independent from the user's actions. I'd much rather provide the user with all the necessary info (full, size, keyUnknown) and prevent an invalid command from altering the table altogether so adding an error output feels redundant. If you feel like it is necessary, it is definitely doable though

doc/base/olo_base_hashtable.md
AnthoJack: No, only _Out_Key_ is updated by this operation. The current key being output is tracked using an independent counter from the one which outputs the value. I guess it could be done to use the same for both though: it is already mentioned in the documentation that no operation should be performed between the fetching of all keys but the only benefit would be one less counter used

The user doesn't know when the end of the table is reached, the *Next Key* operation only finds the next used spot in memory and outputs the key. The way it is meant to be used is described further down: do as many *Next Key* operations as there are elements in the hashtable (given by _Out_Size_). No guarantee is given when it comes to the order the keys will come at because it would require another operation to reset the tracker to the start of the memory. The only guarantee is that using the operation as described without any other operation in-between will output all the stored keys once and once only (I'll add this in the documentation). We can discuss of another way to do this if it's not the way you'd like however

doc/base/olo_base_hashtable.md
AnthoJack: I was skeptical of the use of AXI-stream handshaking on both the input and output but I gave it some thought and it's a really good idea. It should be relatively easy to do on the output and given the kind of handshake I was using before it's a no-brainer to change it to this standard 

When it comes to the groups you suggested, I feel like it could be a bit confusing to new users of the component. Maybe it's only me but when I discover a new component, I like the interfaces to be ordered one of two ways: 
* No groups but in the order they are declared in the entity
* Grouped in the following categories (exact names may change): Synchronisation, data (eventually with in/out subsections), command, status
Ideally, when I design, I make it so the entity declares the interfaces following these groups. I feel like it lets users more easily find the interface they are looking for either by following the way they are declared in the entity or by pinpointing the category it should fall in. Joining the interfaces because they are used together in the function can start being a headache when the features change, get expanded or when an interface can be used by many different operations. I'd rather let the documentation of the different features make it clear which interfaces are used/modified (In which case I admit that I still have some work to do for full and size). Tell me what you think about my reasoning. It's definitely shaped by how I think which may not apply to everyone but I feel like I've found some good points in favor of it which I listed here

Lastly, I don't understand what you mean by "asynchronous" for full and size: They both depend on a counter that tracks the number of pairs stored in the memory and which only changes when an operation is applied which seems like the definition of synchronous (following the clock). Can you explain what you mean by that ?

doc/base/olo_base_hashtable.md
AnthoJack: My plan would be to implement only option 1 for now as I find option 2 hard to justify for now: It adds complexity to the implementation (arguably not a lot since the basic operations needed are already there but still) while hurting performance as it doesn't reduce clustering right away. If someone reaaally needs it in the future, it can always be added but since it's strictly internal, I'm not sure it is necessary

doc/base/olo_base_hashtable.md
AnthoJack: I corrected the formula

TBH, I based this formula on the pseudo-code found on the [Wikipedia page for open addressing](https://en.wikipedia.org/wiki/Open_addressing) and although I understand in which kind of situation each part of it is used, I'm not good at figuring out how (or if) these situations happen so I kinda trusted the pseudo-code and left this part in. I feel like trying to optimise it out may lead to hard to find bugs if it is ever needed

Your example seems correct so I'll add it to the documentation but I'll use it to show compression-first too

doc/base/olo_base_hashtable.md
obruendl: Let me tackle the easiest one first: By "asynchronous status informatin" I mean "not handshaked together with In_Valid/Ready resp. Out_Valid/Ready".  Input and output are "transactions" (with handshwking), the status signals are just a status and can be looked at at any time. I did NOT mean "asynchronous to clock".

Regarding grouping in the documentation: As long as the port naming is indicating the grouping (which now is the case), I am fine to group them differently in the documentation. Probably the order of the _Output Data_ and _Operations_ should be flipped (so "Operations" and "Input Data" are next to each other). But I don't see a reason to change anything on the doc side.

I think the Remaining question now mainly is if the input/request side also gets AXI4-Stream Handshaking or not. Correct?

My opinon regarding this is still that this would be good. 

First of all, it's a common principle open logic follows everywhere (and also documented in the [coding conventions](https://github.com/open-logic/open-logic/blob/main/doc/Conventions.md#axi4-stream-handshaking-signals)). AFAIK users like that everything follows the same handshaking.

Second, it would match the [CAM](https://github.com/open-logic/open-logic/blob/main/doc/base/olo_base_cam.md), which serves a relatively similar purpose. I know that this entity is used and nobody had any complaints or questions to the interface - so it seems OK.

Having the _In_Ready_ signal allows to clearly indicate when the hashtable is ready to a accept the next request, which resolves any questions in the direction of "when can I request the next operation" or "what happens if the next operation is requested before the HashTAble is ready" or "How long do I have to keep _In_Write_ high".

And last but not least, it will simplify the testbenching because AXI4-Stream Verification Components can be used for basically all interfaces (which again is what is done every where in Open Logic and hence also maintenance friendly).

So my opinion regarding the "In_Valid/In_Ready" handshaking is relatively clear. Is there anything that you specifically dislike on the AXI4-Stream handshaking on the "In_xxx" signals?

doc/base/olo_base_hashtable.md
obruendl: Yeah, sure. It's fine to fill it once you are done. 

Some things (like internal architecture, FSM, latency numbers, etc.) have to be done at the end. I 100% agree.

doc/base/olo_base_hashtable.md
obruendl: Sure - You can just state "always '0'" for the ones it is not used. Or "N/A".

doc/base/olo_base_hashtable.md
obruendl: Regarding "Clear after reset": I had the same problem in the CAM. I added a generic "ClearAfterReset_g" there to automatically do this (at the cost of the CAM becoming ready for user operations only after the memory was cleared).

Could we add this to the CAM? It's relatively simple and I feel like this could be valuable for users (one thing less they have to care about).

doc/base/olo_base_hashtable.md
obruendl: I agree, if the "Full" signal is guaranteed to be set at the moment the user could request a write to an already full hash-table, this is sufficient. So we can remove the error signal.

Probably adding an assert to also throw a simulation error if anybody writes to a full hash table would be wide though (so the error at least becomes very prominent in simulation). That's probably more useful than a port anyways.

doc/base/olo_base_hashtable.md
obruendl: OK, so I probably misunderstood something - and I still don't get it. So I need some help here.

My understanding was that this operation searches for the next occupied cell in the RAM and outputs the key stored. Hence I thougt that if we read the memory anyways to get the key - we naturally also would read the value and hence both could be output at zero additional cost (because both are stored together in that RAM cell). 

Did I misunderstand this?

doc/base/olo_base_hashtable.md
obruendl: Let's go the following way: Implement the table with the formula you have and let's then try to apply my optimization. At this point I can explain it to you in more detail. I am pretty sure about this being OK.

But IMO you do not have to boether about it right now. Its a simple optimization we can apply at the end. Just leave the comment open until then.

doc/base/olo_base_hashtable.md
AnthoJack: Keeping this open to remember it once implementation is started

doc/base/olo_base_hashtable.md
AnthoJack: > My understanding was that this operation searches for the next occupied cell in the RAM and outputs the key stored.

That part is correct

However, in my previous implementation, I had separate "pointers" for the RAM-cell whose value is read and the one whose key is read. It was before I decided not to offer any guarantee when not using the key-search feature exactely as described so I thought it would be a good idea to keep them separate so that the user could read a value between 2 key searches and still have correct outputs. Now that the documentation explicitely says that the key-search isn't guaranteed to work if another action is performed before reading ALL the keys, a single pointer can be used as you described so I'll modify the implementation accordingly

Keeping this unresolved so I remember it

doc/base/olo_base_hashtable.md
AnthoJack: I understand how my initial comment may have been misleading: I was initially convinced about using AXI-stream handshaking on the input but skeptical about using it on the output too but I gave it some thought and now I'm fully onboard using it on both sides (and everywhere I can in my following projects)

doc/base/olo_base_hashtable.md
AnthoJack: Yep, I'll add a "ClearAfterReset_g" generic to the hashtable

doc/base/olo_base_hashtable.md
obruendl: Cool. We are good to go for the implementation then.

doc/base/olo_base_hashtable.md
obruendl: Sounds good. The restriction to not allow other operations while reading out keys seems fine to me.

obruendl:
@AnthoJack Looks like we found an agreement on all significant points and youbare good to go for the implementation.

I think I answered all threads where could potentially wait for my answer. Let me know if I missed anything.

obruendl:
@AnthoJack 

**Hashing Function**

I am basically happy with anything beyhond the division only. For me it is okay if you provide one simple option - anybody wanting something more complex can still contribute and add the desired function himself.

You could easily user CRC and just calculate it in a function (instead of instantiating the entity). This should be as simple as adding the loop below into a fuction (more or less):

https://github.com/open-logic/open-logic/blob/6e610f865916303cef55320bac9361d1474f4b29/src/base/vhdl/olo_base_crc.vhd#L129-L141

**Testbenches**

The "not very vunitish" comment seems to be a copy paste error. I will clean this up. All the testbenches that use AXI4-Stream VUnit VCs you can coniser as "good examples".

**Run VUnit**

The the argument passed to run.py is the test-case name. Try adding a asterisks (wildcard) before and after the TB name to include all test-cases of this TB.

`python sim/run.py --nvc *olo_base_hashtable_tb*`

src/base/vhdl/olo_base_hashtable.vhd
obruendl, line 1: Did you already set up the VSG linter? (See [HowTo](https://github.com/open-logic/open-logic/blob/main/doc/HowTo.md#use-the-linter))

I ask because I think some of the formattings (e.g. the UPPER_CASE constants) will be reported by the linter. The earlier you have the linter set-up, the less you have to clean up in the end.

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,324 @@
...
+                    -- EQUATION:
+                    -- R = (rd_idx > wr_idx)((hash >= wr_idx) + (hash > rd_idx)) + 
+                    -- (rd_idx < wr_idx)((hash >= wr_idx) + (hash > rd_idx)) = 
+                    -- A(B+C) + DBC
+                    -- 2 assumptions allow us to simplify it:
+                    -- * rd_idx and wr_idx are never equal -> D = !A
+                    -- * When rd_idx <= wr_idx (!A), hash<rd_idx and hash>wr_idx 
+                    --  cannot exist -> !A!B!C is undefined and can be anything
+                    -- With the 2 previous assumptions, equation can be rewritten
+                    -- R = A(!BC + B!C) + !A(BC + !B!C) = A xor !(B xor C)
```
obruendl: Assumption 1 for simplification already is present in the earlier discussion of the PR.

wr_idx < ird_idx && (hash <= wr_idx || hash > rd_idx) || **rd_idx <= wr_idx** && hash > rd_idx and hash <= wr_idx

Regarding assumption 2:
I don't really see a big simplication. As long the conditions A, B and C are present, it should not matter howo much you simplify the code. The synthesis tool optimizes the logic anyways.

Hence: I suggest to stay with the equation discussed earlier (and state in this comment bove). Better keep the code logical and readable and let the synthesis tool optimize the logic.



wr_idx < ird_idx && (hash <= wr_idx || hash > rd_idx) || rd_idx <= wr_idx && hash > rd_idx and hash <= wr_idx

AnthoJack:
I've finally managed to finish the implementation of both the hashtable and its testbench and everything seems to pass. I've also gone and updated the documentation and added comments to the code. I've decided to drop support for CRC for now and implemented LCG instead which should be a good first algorithm for now

Take a look and tell me if you have anything more to add or modify

obruendl:
@AnthoJack 

Cool thanks. I'll look into it.

What's the pressure level from your end? Do you need it quickly for your work or can I take myelf one or the other week? I am currently teaching which makes my weeks 6 working days, so time is a bit limited in November.

AnthoJack:
No pressure whatsoever

The one we use at work has other modifications that were needed for our use case and uses the old interfaces which I'm not sure I want to update

Bringing it over to open-logic was always a pure spare-time project, sorry if that wasn't clear from the start and resulted in any kind of pressure

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+    use olo.olo_base_pkg_logic.all;
+
+entity olo_base_hashtable is
+    generic (
+        Depth_g : positive;
+        KeyWidth_g : positive;
+        ValueWidth_g : positive;
+        Hash_g : string := "LCG";
+        LcgMult_g : positive := 1103515245; --For LCG. From GCC's implementation
+        LcgIncr_g : positive := 12345; --For LCG. From GCC's implementation
```
obruendl: Can we call these *Hash_LCG_Mult_g* and *Hash_LCG_Incr_g*? I assume every hashing function would come with its settings, so it makes sense to organize the generic names in a clean way.

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+        In_Remove : in std_logic;
+        In_Clear : in std_logic;
+        In_NextKey : in std_logic;
+        In_OpValid : in std_logic;
+        Out_OpReady : out std_logic;
+        Out_DataValid : out std_logic;
+        In_DataReady : in std_logic;
+        Out_KeyUnknown : out std_logic;
+        Out_Full : out std_logic;
+        Out_Pairs : out std_logic_vector(log2ceil(Depth_g) downto 0)
```
obruendl: I few comments on the port names:

1. The convention is `<bundle>_<signal>` - the direction is given by the VHDL`in`/`out`, not by the name. 
    * It should be *In_OpReady* because the signal is related to the *In* bundle
2. Please use simply *_Valid* and *_Ready* (not *_OpValid* or *_DataValid*). That's the convention used everywhere else.
3. Only place signals that are handshaked with the rest of the bundle in the same bundle. *Out_Full* seems to be a static signal and not handshaked with *Out_DataValid* - hence I'd suggest to name this simply *Full*
4. Please split the singals into groups not only in the doc but also in VHDL in the same way, separated by comments.

For the Groups/Bundles, I think we are down to
* Control Signals (Clk/Rst)
* In (or Op in case you prefer)
* Out
* Status (I only see Full here)

obruendl: I think this should be `log2ceil(Depth_g+1)-1 downto 0` right?

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+        Out_DataValid <= '0';
+        Hash_InKey <= Ram_RdData.key;
+
+        --FSM
+        case reg_sp.ht_state is
+            when IDLE =>
+                --Hashtable ready for new operation
+                Out_OpReady <= '1';
+                if In_OpValid <= '1' then --AXIS handshake
+                    reg_sn.user_data <= (In_Key, In_Value, '1'); --Memorise input data
```
obruendl: I'd suggest to add an `assert` here to check if really only one of the operation requests is active. If `In_Write` and `In_Read` are set at the same time, only the write is executed. This is not something a user should do and it might lead t oconfusing errors, hence an assert would be good.

Plase do `-- synthesis off` and `-- synthesis on` around it (some tools like Yosys don't like asserts in synthesizable code).

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+            when SEARCH_INIT =>
+                --Init search at index given by hash
+                Hash_InKey <= reg_sp.user_data.key;
+                reg_sn.key_found <= '0';
+                Ram_RdAddr <= Hash_Out;
+                Ram_RdEna <= '1';
+                reg_sn.cnt <= to_unsigned(0, reg_sn.cnt'length);
+                reg_sn.rd_idx <= Hash_Out;
+                reg_sn.ht_state <= SEARCH_KEY;
+
```
obruendl: Can we move `SEARCH_KEY` here? It would help readability if the two search states are next to each other in the order they are executed in.

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+        RamStyle_g => RamStyle_g,
+        RamBehavior_g => RamBehavior_g
+    )
+     port map(
+        Clk => Clk,
+        Wr_Addr => std_logic_vector(Ram_WrAddr),
+        Wr_Ena => Ram_WrEna,
+        Wr_Data => to_vector(Ram_WrData),
+        Rd_Addr => std_logic_vector(Ram_RdAddr),
+        Rd_Ena => Ram_RdEna,
```
obruendl: Is there a reason to not keep `Rd_Ena` always asserted?

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+                    reg_sn.ht_state <= IDLE;
+                end if;
+
+            when CLUSTER_COMP =>
+                --Cluster compression
+                Ram_WrAddr <= reg_sp.wr_idx;
+                if reg_sp.key_found = '1' then
+                    if Ram_RdData.used <= '0' then --Done
+                        Ram_WrEna <= '1';
+                        reg_sn.ht_state <= REMOVE;
```
obruendl: To my understanding teh cluster compression is done before the REMOVE. Shouldn't this be the other way round?

Remove basically creates a whole in the cluster and this can be compressed immediately.

At least this would be my understanding.

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+        if ClearAfterReset_g then
+            ResetVal.ht_state <= CLEAR;
+        end if;
+        
+    end process;
+
+    --Register process
+    process (Clk, Rst)
+    begin
+        if Rst = '1' then
```
obruendl: Please only use synchronous resets. This is a convention of open-logic for good reasons (asynchronous resets have many dangerous caveats).

Please follow the Reset pattern from other logic blocks. Having the registered process in the same form everywhere helps with readability.

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+            reg_sp <= ResetVal;
```
obruendl: Please only reset the regsiters that really require a reset.

For example `reg_sn.user_data` does not require a reset because it is always set before it is read (it is set when the operation is detected and read only in subsequent states).

Reset Fanout is a timing problem, therefore we should not blindly reset all regsiters. At least this is how Open Logic does.

src/base/vhdl/olo_base_pkg_math.vhd
```diff
@@ -654,5 +657,17 @@ package body olo_base_pkg_math is
 
         return Min_v;
     end function;
+    
+    -- *** Calculate next pseudorandom value for Linear Congruential Generator ***
+    function lcg_prng(a : unsigned; mult, incr : integer) return unsigned is
```
obruendl: Would it make sense to pass the width of the output (and truncate the output accordingly)?

obruendl: please follow the Naming convention. In this case `lcgPrng`

obruendl: Documentation and TB for this function are missing. Please add.

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
obruendl, line 1: Can we write the TB in prover VUnit style (transaction level)?

This would mean using an _axi_stream_master_t_ to send in the commands, an _axi_stream_slave_t_ to retrieve the reseponse and a control process that defines the test cases. Not only would this make the TB more readable but also would the code become much more concise.

You can refer to the other TBs as example. For example _olo_base_cam_ has such a TB wih AXI4-Stream VCs on input and output side. The _olo_base_cam_ TB is a good orientation anyways because the entities are somewhat related.

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
```diff
@@ -0,0 +1,404 @@
...
+    use vunit_lib.check_pkg.all;
+
+library olo;
+    use olo.olo_base_pkg_math.all;
+    use olo.olo_base_pkg_logic.all;
+
+entity olo_base_hashtable_tb is
+    generic (
+        runner_cfg : string;
+        Depth_g : positive := 8;
```
obruendl: For the TB, I think it is fine to make some generics constant:

Depth_g, KeyWidth_g, ValueWidth_g, LcgMult_g, LcgIncr_g as well as the Ram generics can be made constant.

For the generics you keep, please add some different generic configurations to reach good coverage to the run.py.

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,361 @@
...
+        --Default values
+        reg_sn <= reg_sp; --Keep current register values by default
+        Out_OpReady <= '0';
+        Ram_RdEna <= '0';
+        Ram_RdAddr <= (others => '0');
+        Ram_WrAddr <= (others => '0');
+        Ram_WrEna <= '0';
+        Ram_WrData <= Ram_RdData; --Write data is read data by default
+        Out_DataValid <= '0';
+        Hash_InKey <= Ram_RdData.key;
```
obruendl: There is a combinatorial path containing the RAM read logic, the hashing function (which includes the multiplier for LCG) and goes back to the RAM address. This is too much logic (it restricts the clock frequency significantly).

Looking at this, I wanted to check the reason for not implementing CRC as hashing. This would avoid the multiplier and therefore run much faster I think.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,210 @@
...
+| LcgIncr_g         | positive  | 12345     | Multiplier used in the LCG (Default value from libc) |
+| RamStyle_g        | string   | "auto"  | Passed to [*olo_base_ram_sdp*](./olo_base_ram_sdp.md). Refer to the documentation of this component for more info |
+| RamBehavior_g     | string   | "RBW"   | Passed to [*olo_base_ram_sdp*](./olo_base_ram_sdp.md). Refer to the documentation of this component for more info  |
+| ClearAfterReset_g | boolean | true | Clear memory after a reset to prevent erroneous values |
+
+Current supported hash algorithms are:
+* *DIVISION*: Key's value is used as-is, except for modulo against depth of hashtable to obtain a valid index
+* *LCG*: [Linear Congruential Generator](https://en.wikipedia.org/wiki/Linear_congruential_generator) PRNG algorithm is used. The multiplier and increment for the algorithm can be chosen using _LcgMult_g_ and _LcgIncr_g_ respectively. Modulus is width of the hastable's indices
+
+## Interfaces
```
obruendl, line 37: Please organize the ports into exactly the bundles we have in HDL. Including the handshaking for every bundle in the corresponding table.

This might be a matter of personal style and I don't want to say separating the handshaking into one section is _wrong_ or what I suggest is _more right_. My point here is that it should be documented in the same form as it is done for all other entities so it is _consistent_

obruendl (commented):
I have two points regarding tooling:

The code currently throws linting errors - which normal in early stages. Those should be fixed. Do you have the VSG linter installed or do you need help? Generally the setup should be documented.

The TB currently does not achieve 100% branch and statement coverage yet. Again I wanted to check if you have questa installed and you know how to measure coverage or if you need any help.

Generally I think the TB should have some more test cases. This includes filling the table completely, different handshaking timing, etc.

src/base/vhdl/olo_base_hashtable.vhd
AnthoJack: I setup the linter and corrected all mistakes it was throwing

src/base/vhdl/olo_base_hashtable.vhd
AnthoJack: Fine to me

src/base/vhdl/olo_base_hashtable.vhd
AnthoJack: I'll rename the In_* and Out_* signals according to the bundles you suggested (Op would be fine if the data was bundled elsewhere but In seems fine since it isn't)

Full and Pairs are both static (not handshaken) and full is combinatorially linked to pairs (result of a comparison) so I'd group them together

AnthoJack: No because if it were, it would be impossible to store the maximum amount when hashtable is full. For example, if the depth is 8, pairs should be able to return values 0(empty) -> 8(full). log2ceil(8) is 3 so std_logic_vector(2 downto 0) would only be able to store values 0->7

Now that I'm thinking about it, instead of a comparison, I could actually just return Out_Pairs MSB into Out_Full. I don't know if synthesizers are able to perform that kind of optimisation automatically

src/base/vhdl/olo_base_hashtable.vhd
AnthoJack: I tried modifying it to keep it set but my tests failed and I'm not sure I want to modify that now that I went through the hassle of making everything work with it only set when needed. If it's meant to be always set, why create the ram with this input in the first place ? Now that it's there might as well use it properly

src/base/vhdl/olo_base_hashtable.vhd
AnthoJack: Doing cluster compression first allows us to do only one remove. Doing the remove makes a hole that may be filled by the cluster compression and would force us to do another remove after cluster compression

src/base/vhdl/olo_base_hashtable.vhd
AnthoJack: I'll fix this combinatorial path but that will probably mean one more state for cluster compression . Where do you think the split should happen ? After the read logic or after the hashing ?

When it comes to CRC, everytime I try to tinker with it I feel like I'm wandering in the dark and I don't like that so I'm focusing on making the hashtable work for now. Once the hashtable is almost ready to be merged, I'll try again to add the CRC. after all, adding another algorithm should be as simple as adding it to the if-generate statement and running the test bench with a config that uses it. The implementation of the algorithm is what I'm struggling with right now so I'm postponing it till later

obruendl: I should have time to look into the changes in the coming two days.

I could add CRC if it helps you. Just give me push permissions to the branch. Shall I?

src/base/vhdl/olo_base_hashtable.vhd
obruendl: True - you are right.

src/base/vhdl/olo_base_hashtable.vhd
AnthoJack: I sent you an invite. Tell me back if it doesn't give you push permission. I'd be very grateful if you could implement CRC

AnthoJack: I sent you an invite. Tell me back if it doesn't give you push permission. I'd be very grateful if you could implement CRC

AnthoJack:
Hey
Thank you for the implementation. I'm a bit busy these days but I fully intend to work on the hashtable during the holidays so I'll fix the TB and check everything you mentioned. Have a nice end of year

doc/base/olo_base_hashtable.md
AnthoJack: Done

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
AnthoJack: I looked at the other testbenches but found no way to automatically produce different generic configuration sets. Do you know of any way to do it ?

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
AnthoJack: Done

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
obruendl: @AnthoJack 

Sure. The generic configurations are generated in python script that runs the tests (that's standard in VUnit).

Have a look at /sim/test_configs/olo_base.py

The file contains one short section for every simualted entity and for almost all of them I generate different generic sets. Please do not go over board - do generate sets to cover all cases but do not generate "all possible combinations" (if we'd do this for every entity, simulation times explode).

Ifyou have modelsim/questa you can check simulation coverage by running

```
cd sim
python3 run.py --modelsim --coverage *hashtable*
python3 AnalyzeCoverage.py
```

Once all code is covered we probably are okay.

AnthoJack:
> @AnthoJack 99% of the issues with `wait until <condition>` statements are because the signal is already in that condition before the statement is executed. The solution is to add an if statement before:
> 
> ```vhdl
> if not <condition> then
>     wait until <condition>
> end if;
> ```

That fixed it, thanks. Seems like a weird thing not to use `wait until rising_edge(<signal>)` to expect the signal to change but I'll keep that in mind

obruendl:
@AnthoJack 

From the quick back and forth you had with @tasgomes , I conclude you are working on it.

What's the state? Still WIP or is it already ready for the next review from me?

AnthoJack:
> @AnthoJack 
> 
> From the quick back and forth you had with @tasgomes , I conclude you are working on it.
> 
> What's the state? Still WIP or is it already ready for the next review from me?

It's ready for your next review. Right now I'm working on trying to make coverage analysis work with NVC since I don't have access to Questa/modelsim

AnthoJack:
Just now noticed that vunit has already implemented support for coverage analysis and merging using NVC in their repo. Sadly, it's not currently in a released version of the vunit_hdl module for python. If you need more modifications, I'll try to use the implementation on their GitHub to make it work here temporarily so that I don't have to rely on you everytime I want to check coverage but if you are satisfied with the current state of the branch I'll just wait for a version of vunit_hdl to support it

obruendl:
@AnthoJack 

Regarding NVC coverage: You can merge in the branch "feature/nvc-coverage". For this to work you need:
- VUnit 5.0.0.dev7
- NVC: https://github.com/nickg/nvc/commit/f880ffcd16d18bf0a1353dce48bcd9108d6744c1
  - Really this one (or something newer). This commit contains a bugfix to an issue I opened.
  
 You can then generate a coverage report using:
 ```
 python3 run.py *hashtable* --nvc --coverage
 ```
 
 I'll look into the code in the coming days (4.5.0 is out, so I should have time to come back to it). I can also run coverage with modelsim.

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,452 @@
+---------------------------------------------------------------------------------------------------
+-- Copyright (c) 2026 by Oliver Bruendler
+-- Authors: Oliver Bruendler
```
obruendl, line 3: @AnthoJack add your name here (before mine, you did majority of the work)

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
```diff
@@ -0,0 +1,492 @@
+---------------------------------------------------------------------------------------------------
+-- Copyright (c) 2026 by Oliver Bruendler, Switzerland
+-- Authors: Oliver Bruendler
```
obruendl, line 3: @AnthoJack add your name here (before mine, you did majority of the work)

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,204 @@
...
+### Axis Handshaking
+
+| Name      | In/Out | Length        | Description                                  |
+| :-------  | :----- | :--------     | :------------------------------------------- |
+| In_Ready  | out    | 1             | Axi-Stream handshaking signal that hashtable is ready to accept an operation |
+| In_Valid  | in     | 1             | Axi_Stream handshaking signal that an operation is requested |
+| Out_Ready | in     | 1             | Axi-Stream handshaking signal that user is ready to accept data |
+| Out_Valid | out    | 1             | Axi_Stream handshaking signal that hashtable has data for the user |
+
+### Input
```
obruendl, line 55: This is the ports description - can we move the latency documentation to the section _Architecture > Operations_?

Reason: Port descritpions really only should be port descriptions to stay consistent with all other elements in Open Logic.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,204 @@
...
+
+Latency of *cluster_comp* is: *1 + (cluster_size - element_position_in_cluster)*
+
+### Output
+
+| Name      | In/Out | Length           | Default | Description                                   |
+| :-------- | :----- | :--------        | ------- | :-------------------------------------------- |
+| Out_Key   | out    | _KeyWidth_g_      | N/A     | Stored key output |
+| Out_Value | out    | _ValueWidth_g_    | N/A     | Output value                                  |
+| Out_KeyUnknown    | out    | 1                | N/A     | _In_Key_ does not exist in the hashtable (WR/RM/RD operations)  |
```
obruendl, line 79: I think we discussed the timing of this signal before.

For me it is unclear when the signal is asserted. As stated before, I really believe it should be as follows.
- _Out_KeyUnknown_ shall be handshaked with _Out_Valid_ and _Out_Ready_
- Every command shall produce a _Out_Xxx_ response (_Out_Valid_/_Out_Ready_ handshaking cycle) 

This also will help the TB. You can then check the _Out_KeyUnknown_ with the AxiS-Slave like the rest of the output.

Alternatively, if you really do not want to have _Out_KeyUnknown_ handshaked with _Out_Valid_/_Out_Ready_, it should not have the _Out_ prefix. In that case I'd vote for a name _Status_LastCmdKeyUnknown_ or the like and you should provide waveforms to describe when the _KeyUnknown_ is updated. There should be at least two waveforms, one for a command without _Out_Valid_ (e.g. a write) and one with _Out_Valid_ (e.g. a read).

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,204 @@
...
+| Status_Pairs          | out    | _log2ceil(Depth_g)_+1 | N/A     | Number of Key-Value pairs stored |
+
+
+## Architecture
+
+### Notes
+
+* Operations requested while hashtable isn't ready (_Out_Ready_ = '1') are ignored
+* A supplementary `used` bit is stored in memory alongside each key-value pair to indicate which memory words are occupied (`used` = 1). The total width of the internal memory is thus: RamWidth = KeyWidth_g + ValueWidth_g + 1
+* If the used *olo_base_sdp* is in an unknown state after reset, the memory must be cleared before accepting any operation to prevent the hashtable from finding erroneous `used` bits set to 1 when the hashtable should be empty. The
```
obruendl, line 96: There is a "The" at the end of the line. Looks like an incomplete sentence - please check.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,204 @@
...
+* If the used *olo_base_sdp* is in an unknown state after reset, the memory must be cleared before accepting any operation to prevent the hashtable from finding erroneous `used` bits set to 1 when the hashtable should be empty. The 
+
+### Operations
+
+#### Key search
+
+The hastable uses a hash function to obtain a memory index from the provided key. As the hashing could produce the same result for different keys,the hashtable uses [linear open-addressing](https://en.wikipedia.org/wiki/Open_addressing) to resolve those collisions: the following elements are checked until either the key or an empty spot is found, creating clusters (or search-chains). When a key is found, _Out_KeyUnknown_ is reset to '0' and set to '1' otherwise. A counter is also used to prevent endless looping when memory is full as there would be no empty spot to end the search on
+
+The more pairs are stored, the more chances of collisions and thus the formation of clusters whose size rapidly reduces the performance of the hashtable. When performance is an important factor, the hashtable should be dimensioned such that it never comes close to being full
```
obruendl, line 104: I think there are recommendations how much a hash-table can be filled without significant performance degradation (I think I remember something like 25%). Can you probably check and add a link or a note? For users not deep into hash tables this could be very helpful.

doc/base/olo_base_hashtable.md
```diff
@@ -0,0 +1,204 @@
...
+
+Note also that memory is used cyclically and thus, elements at indices 7 and 0 would be considered as part of the same cluster. When reconstituting the cluster, the following equation is used to determine if an element's hash places it outside the search chain (cluster broken by remove):
+
+wr_idx < rd_idx && (hash <= wr_idx || hash > rd_idx) || rd_idx <= wr_idx && hash > rd_idx && hash <= wr_idx
+
+Where `wr_idx` is the index of the current "hole" in the cluster, `rd_idx` is a following element of the cluster and `hash` is the hash of the element at `rd_idx`. When the equation is true, the element at `rd_idx` is copied at `wr_idx`. Once an empty element is found, the cluster is complete and the last copied element is emptied. Here's a more complete example using compression-first as explained above:
+
+Idx | Init  | Compression | Removal | Cmp | Rm | Cmp | Rm | Cmp | Rm | Cmp | Rm
```
obruendl, line 183: I like this table.

A few things are missing to make easy for the reader to understand:

1. Explain the notation A(4) means Hash=4 and Value=A if I am not mistaken
2. Group the Compress/Remove pairs (name them like "Compression-1", "Removal-1" and "Cmp-2", "Rm-2" etc. and then describe which element is removed in which step (e.g. Cmp-2 and Rm-2 remove element A(4) from Idx=4).

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
```diff
@@ -0,0 +1,492 @@
...
+        signal net : inout network_t;
+        checkValue : std_logic_vector(ValueWidth_c-1 downto 0);
+        checkMsg   : string) is
+        variable StreamData_v   : std_logic_vector(ValueWidth_c+KeyWidth_c-1 downto 0);
+        variable ReadValue_v    : std_logic_vector(ValueWidth_c-1 downto 0);
+        variable TuserDiscard_v : std_logic;
+    begin
+        pop_axi_stream(net, AxisSlave_c, StreamData_v, TuserDiscard_v);
+        ReadValue_v := StreamData_v(ValueWidth_c-1 downto 0);
+        check_equal(ReadValue_v, checkValue, checkMsg);
```
obruendl, line 177: Is there any reason for making this operation blocking instead of using _check_axi_stream()_?

I think it would be easier for writing the test-cases if you could simply provide all commands and all expected outputs and then let the VCs check them in order instead of having a blocking htReadCheck().

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
```diff
@@ -0,0 +1,492 @@
...
+    use osvvm.RandomPkg.all;
+
+library olo;
+    use olo.olo_base_pkg_math.all;
+    use olo.olo_base_pkg_logic.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity
+---------------------------------------------------------------------------------------------------
+entity olo_base_hashtable_tb is
```
obruendl, line 29: Can you make it work with

```
-- vunit: run_all_in_same_sim
entity ...
```

Questa is terribly slow in starting, so I always run all test-cases in one simulation. It's the same for all entities and I'd like to keep it consistent.

I think the only reason for it to fail is that the memory is not cleared when _ClearAfterReset_g=fasle_ (exactly as expected). In that case I think it'd be fine if you just add a manual "Clear" command between the reset and the first command of the normal test sequence.

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
```diff
@@ -0,0 +1,492 @@
...
+    generic (
+        runner_cfg        : string;
+        Hash_g            : string  := "CRC32";
+        ClearAfterReset_g : boolean := true
```
obruendl, line 33: For the case _ClearAfterReset_g_, do we test if the hash table state is consistent and working (i.e. the old content really is preserved) afte reset?

I suspect the hashtable may be in a inconsistent state after reset because the "Pairs" counter is reset but the memory content is not.

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
```diff
@@ -0,0 +1,492 @@
...
+                    check(Status_Busy = '1', "Hashtable status clearing after reset");
+                else
+                    check(In_Ready = '1', "Hashtable AXIS ready after reset");
+                    check(Status_Busy = '0', "Hashtable status ready after reset");
+                end if;
+                check(Out_Valid = '0', "No data output after reset");
+                check(Status_Full = '0', "Hashtable not full after reset");
+                check_equal(Status_Pairs, 0, "Hashtable empty after reset");
+
+            elsif run("TestFeatures") then
```
obruendl, line 244: Can you break-up the "TestFeatures" case into individual test cases? I'd expect at least:

- Write one pair then read it
- Write a handful of pairs, then read them
- Write pairs with a collision (same hash), then read them.
  - The collision at least for Hash_g="MODULO" is easy to provoke
- Write some pairs, remove one, read pairs that are in there and verify the one that was removed is gone
- Write some pairs, clear, check if all pairs are gone
- Write some pairs, read them through "NextKey"
- Hash table going full
- Cluster compression around the wraparound (when the element at the highest address Depth_g-1, Depth_g-2, 0 or 1 is removed - try all of them). AGain for Hash_g="MODULO" this is easy to control.
- Push commands into the hash table back-to-back or with some cycles in between (I think you can run the same test cases and control the delay between commands through a generic)

src/base/vhdl/olo_base_hashtable.vhd
```diff
@@ -0,0 +1,452 @@
...
+    -- Verification asserts
+    -- Depth is power of 2
+    assert log2(Depth_g) = log2ceil(Depth_g)
+        report "###ERROR###: olo_base_hashtable - Depth_g must be a power of 2";
+    -- Width of key must be bigger than width of indices to prevent memory underuse. Depth_g twice as big as number of keys tolerated to avoid clustering
+    assert KeyWidth_g+1 >= PairsIdx_c
+        report "olo_base_hashtable - Memory underuse over 2x: Not enough different key values to fill half the hashtable"
+        severity warning;
+    -- Check Hash Function
+    assert Hash_c = "CRC32" or Hash_c = "MODULO"
```
obruendl, line 180: I switched to compare all string generics case-insensitively (there is a _compareNoCase()_ function in _olo_base_pkg_string_).

Please update accordingly.

This implies you merge _develop_ into your feature branch - which is required anyways to resolve merge conflicts.

test/base/olo_base_hashtable/olo_base_hashtable_tb.vhd
```diff
@@ -0,0 +1,492 @@
...
+
+library osvvm;
+    use osvvm.RandomPkg.all;
+
+library olo;
+    use olo.olo_base_pkg_math.all;
+    use olo.olo_base_pkg_logic.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity
```
obruendl, line 27: I checked code coverage. I added the uncovered parts I found. Please extend the testbench accordingly to cover those as well.

I think especially with "Hash_g=MODULO" you have good control to exercise code in detail.

Snapshot 1:

Image

Snapshot 2:

Image

To run it yourselves: You can download the free questa simulator version from altera. This is the same I use (it can do coverage). It works perfectly fine for open logic purposes.

obruendl (commented):
I think we are getting closer.

A few significant changes I applied:
- I removed LCG becauseof its poot timing performance (multiply add in one cycle). I think with CRC32 being added, there is no benefit for LCG. Better not giving users an option to produce poor timing.
- I replace the LCG random number generation in the TB and replaced it by normal OSVVM random number generation.
- I cleaned up all inter issues in production code as well as testbench
- I added a bit of additional structure (block comments) to match the other Open Logic code

I think production code looks now pretty OK. 

The documentation as well - one or the other waveform would help though.

What requires still some touches is mainly the testbench.

AnthoJack:
Hey @obruendl 

Just a message to say that the fact that I haven't touched this PR in 3 months doesn't mean I gave up on it. Life has just been frankly all over the place recently (new job, apartment search, paperwork...) so I haven't found the time and energy to both keep my life together and work on side projects. I should have a new apartment and some time off next month so I'll try to get back to it then

Thanks for your patience

PR #236, opened 2025-10-29
FEATURE: olo_base_rate_limit

obruendl:
When thinking about fixed-point entities like FIR filters, I thought the addition of a rate limiter could be a good idea.

Many DSP components (whether provided by Open Logic or any other source) have rate limitations and it would be good allowing users to simply limit the input rate into the processing chain instead of having to handle ready along the full chain. I therfore wrote down my idea for a rate limiter.

### DIFF
--- a/src/base/vhdl/olo_base_rate_limit.vhd
+++ b/src/base/vhdl/olo_base_rate_limit.vhd
@@ -154,7 +154,7 @@ begin
         p_comb : process (all) is
             variable v                : TwoProcess_r;
             variable OutputTransfer_v : boolean;
-            variable PeriodMin1_v     : natural;
+            variable PeriodMin1_v     : natural; -- Period minus 1
             variable MaxSamples_v     : natural;
             variable SmoothLimit_v    : natural;
         begin
@@ -165,10 +165,10 @@ begin
             if RuntimeCfg_g then
                 v.CfgPeriod     := to01(Cfg_Period);
                 v.CfgMaxSamples := to01(Cfg_MaxSamples);
-                -- to01 and resize required to workaroudn simulation issues. Functionally they do not
+                -- to01 and resize required to workaround simulation issues. Functionally they do not
                 -- have impact so this is tolerable.
                 v.CfgSmoothLimit := unsigned(to01(Cfg_Period)) - resize(unsigned(to01(Cfg_MaxSamples)), Cfg_Period'length);
-                PeriodMin1_v     := to_integer(unsigned(to01(r.CfgPeriod)));
+                PeriodMin1_v     := to_integer(unsigned(to01(r.CfgPeriod))); -- This is period minus 1 because the definition of the port is like that
                 MaxSamples_v     := to_integer(unsigned(to01(r.CfgMaxSamples)))+1;
                 SmoothLimit_v    := to_integer(r.CfgSmoothLimit);
                 -- synthesis translate_off


doc/base/olo_base_rate_limit.md
```diff
@@ -0,0 +1,86 @@
...
+Note how the samples never exceed the rate of one sample every two clock cycles. Also not over short periods of time.
+
+## Generics
+
+| Name            | Type      | Default  | Description                                                  |
+| :-------------- | :-------- | -------- | :----------------------------------------------------------- |
+| Width_g         | positive  | -        | Width of _In_Data_ and _Out_Data_                            |
+| RegisterReady_g | boolean   | true     | If true, _In_Ready_ is registered to improve timing          |
+| Mode_g          | string    | "SMOOTH" | Rate limiting mode, either "BLOCK" or "SMOOTH"               |
+| Period_g        | time      | 1 ms     | Time period for rate limiting in clock cycles                |
```
rbrglez: Change the type to natural

doc/base/olo_base_rate_limit.md
```diff
@@ -0,0 +1,86 @@
...
+- _BLOCK_: In this mode the component does not forward more than a limited number of samples over a given period but
+  still does forward short bursts at full speed as long as the average rate is not exceeded.
+- _SMOOTH_: In this mode the component spaces samples evenly over time to achieve a constant output data rate.
+
+### _BLOCK_ Mode
+
+The sample figure is drawn for _Period_g_=4 clock cycles and _MaxSamples_g_=2 samples per period. Periods on the output
+side are marked in colors.
+
+![wave_smooth](misc/olo_base_rate_limit_block.png)
```
rbrglez: rename to wave_block

doc/base/olo_base_rate_limit.md
```diff
@@ -0,0 +1,86 @@
...
+next block (turquoise).
+
+### _SMOOTH_ Mode
+
+The sample figure is drawn for _Period_g_=4 clock cycles and _MaxSamples_g_=2 samples per period. Periods on the output
+side are marked in colors.
+
+![wave_smooth](misc/olo_base_rate_limit_smooth.png)
+
+Note how the samples never exceed the rate of one sample every two clock cycles. Also not over short periods of time.
```
rbrglez: > Also not over short periods of time.

What did you mean with this sentence?

Are you referring to the beginning of the waveform, where In_Valid remains high for multiple clock cycles, but the output handshake occurs only once every two clock cycles?

doc/base/olo_base_rate_limit.md
obruendl: fixed

doc/base/olo_base_rate_limit.md
obruendl: fixed. I chose "positive" because a value of zero does not make any sense.

doc/base/misc/olo_base_rate_limit_block.png
rbrglez, line 1: Why is data 5,6,7 and 8 handshaked on the output at full speed?
Is it because In_Valid has changed from low to high at the start of data(5) handshake?

rbrglez, line 1: I’d like to see more of this unit’s behavior starting from data(5) onward, with In_Valid held high throughout. I’m particularly interested in whether, after the initial burst of four handshaked data values (5, 6, 7, and 8), the unit stabilizes and begins performing output handshakes at the expected rate of two handshakes every four clock cycles.

rbrglez, line 1: Are the waveforms correct, that there is no delay between input handshake and the output handshake?

doc/base/olo_base_rate_limit.md
```diff
@@ -0,0 +1,86 @@
...
+
+### _BLOCK_ Mode
+
+The sample figure is drawn for _Period_g_=4 clock cycles and _MaxSamples_g_=2 samples per period. Periods on the output
+side are marked in colors.
+
+![wave_block](misc/olo_base_rate_limit_block.png)
+
+Note how the rate can exceed the limit if short bursts occur at the end of one block (blue) and the beginning of the
+next block (turquoise).
```
rbrglez: Can you explain why this didn't happen for yellow(1,2) and orange(3,4)) data?

doc/base/misc/olo_base_rate_limit_block.json
rbrglez, line 1: Here’s a quick and dirty experiment I did with wavedrom to make the periods visible. Feel free to ignore it if it’s not useful.

```

{signal: [
  {name: 'Clk',          wave: 'P.................'},  
  {name: 'In_Valid',     wave: '1.....0.1........0'},
  {name: 'In_Ready',     wave: '1.0.1.0.1...0...10'},
  {name: 'In_Data',      wave: '222..2x.22222....x', data:["1", "2", "3", "4", "5", "6", "7", "8", "9"]},
  {name: 'Out_Valid',    wave: '1.0.1.0.1...0.1..0'},
  {name: 'Out_Ready',    wave: '1............0..10'},
  {name: 'Out_Data',     wave: '33x.44x.5566x...7x', data:["1", "2", "3", "4", "5", "6", "7", "8", "9"]},
  { node: '0...1...2...3...4'},
],
  edge: ['0<->1 Period 1', '1<->2 Period 2', '2<->3 Period 3', '3<->4 Period 4']
  }
 ```

**Comment by obruendl** [comment]
@rbrglez 

Review comments addressed. Please check if things are clear now.

Again - the review added value. Thanks.

File: doc/base/olo_base_rate_limit.md [hunk 5]
```diff
@@ -0,0 +1,88 @@
...
+
+## Generics
+
+| Name            | Type      | Default  | Description                                                  |
+| :-------------- | :-------- | -------- | :----------------------------------------------------------- |
+| Width_g         | positive  | -        | Width of _In_Data_ and _Out_Data_                            |
+| RegisterReady_g | boolean   | true     | If true, _In_Ready_ is registered to improve timing          |
+| Mode_g          | string    | "SMOOTH" | Rate limiting mode, either "BLOCK" or "SMOOTH"               |
+| Period_g        | positive  | -    | Time period for rate limiting in clock cycles                |
+| MaxSamples_g    | positive  | 1      | Maximum number of samples allowed per _Period_g_             |
```
rbrglez: @obruendl , could you make Period_g and MaxSamples_g be runtime configurable from e.g. registers?

obruendl: I had the same thought but decided to keep things simple.

Reasons:
- I'd assume this entity mostly be used to avoid overrunning downstream components with fixed processing capabilities. This usually will be compile-time defined I would think. Let me know if you see real use cases that require runtime configurability.
- I recently learned that port default values don't work when instantiating entities from Verilog. Hence veriloggers would always have to assign those ports.

src/base/vhdl/olo_base_rate_limit.vhd
```diff
@@ -0,0 +1,220 @@
...
+
+begin
+
+    -- *** Assertions ***
+    assert MaxSamples_g <= Period_g
+        report "olo_base_rate_limit: MaxSamples_g (" & integer'image(MaxSamples_g) &
+               ") must be <= Period_g (" & integer'image(Period_g) & ")"
+        severity failure;
+
+    assert Mode_g = "SMOOTH" or Mode_g = "BLOCK"
```
rbrglez: This check should be case-insensitive. Mode_g could be "smooth", "Block", ...
A few other components in Open-Logic already perform this check in a case-insensitive way.

sim/test_configs/olo_base.py
```diff
@@ -343,3 +343,23 @@ def add_configs(olo_tb):
...
+        for Mode in ["SMOOTH", "BLOCK"]:
+            named_config(tb, {'Mode_g': Mode, 'Period_g': Period, 'MaxSamples_g': MaxSamples})
+    # Test different RegisterReady settings
+    for RegisterReady in [True, False]:
+        named_config(tb, {'RegisterReady_g': RegisterReady})
+    # Test with random stall enabled
+    for RandomStall in [True, False]:
+        named_config(tb, {'RandomStall_g': RandomStall})
+    # Test different data widths
+    for Width in [8, 16]:
```
rbrglez: There is a duplicate test for `Width = 16`. This case is already covered earlier:

When `Mode = "BLOCK"`

```
for Mode in ["SMOOTH", "BLOCK"]:
    named_config(tb, {'Mode_g': Mode})
```

And here

When `RandomStall = False`

``` 
for RandomStall in [True, False]:
    named_config(tb, {'RandomStall_g': RandomStall})
```
        
Please review the remaining tests and ensure that no identical test cases are being executed multiple times.

src/base/vhdl/olo_base_rate_limit.vhd
```diff
@@ -0,0 +1,220 @@
...
+                r <= r_next;
+                if Rst = '1' then
+                    r.SmoothCounter  <= (others => '0');
+                    r.PeriodCounter  <= 0;
+                    r.SamplesCounter <= 0;
+                end if;
+            end if;
+        end process;
+
+        -- Gatged handshaking signals
```
rbrglez: fix typo
Gated

rbrglez (commented):
Apart from few minor nitpicks, code looks good

sim/test_configs/olo_base.py
obruendl: Fixed. Please review and resolve if you agree.

src/base/vhdl/olo_base_rate_limit.vhd
obruendl: Fixed. Please review and resolve if you agree.

obruendl:
@rbrglez 

I added runtime configurability. Please review again.

sim/test_configs/olo_base.py
```diff
@@ -343,3 +343,18 @@ def add_configs(olo_tb):
...
+                  'RandomStall_g': False,
+                  'Width_g': 16}
+    # Test different Period/MaxSamples combinations
+    for Period, MaxSamples in [(5, 5), (5, 3), (5, 1), (1, 1)]:
+        for Mode in ["SMOOTH", "BLOCK"]:
+            named_config(tb,tb_configs | {'Mode_g': Mode, 'Period_g': Period, 'MaxSamples_g': MaxSamples, 'RandomStall_g': True})
+            named_config(tb,tb_configs | {'Mode_g': Mode, 'Period_g': Period, 'MaxSamples_g': MaxSamples, 'RandomStall_g': False})
+    # Test a different value for static settings
+    named_config(tb,tb_configs | {'RegisterReady_g': False, 'Width_g': 8, 'RuntimeCfg_g' : False, 'Mode_g' : "SMOOTH"})
+    named_config(tb,tb_configs | {'RegisterReady_g': False, 'Width_g': 8, 'RuntimeCfg_g' : False, 'Mode_g' : "BLOCK"})
```
rbrglez, line 360: NITPICK:
Right now, the code relies on the default values of the generics defined in the testbench olo_base_rate_limit_tb.vhd. I’d prefer not to have to check the testbench just to understand which generics are being used in the test configuration. Ideally, all information about the generics should be explicitly defined here in olo_base.py.

This is admittedly a minor point, but making the generics explicit in one place would prevent issues if the default values in the testbench ever change. It would also make the setup easier to understand, since you’d only need to look at a single file (olo_base.py) to know exactly what generics are being used.

src/base/vhdl/olo_base_rate_limit.vhd
```diff
@@ -0,0 +1,260 @@
...
+        In_Data         : in    std_logic_vector(Width_g-1 downto 0);
+        In_Valid        : in    std_logic                                           := '1';
+        In_Ready        : out   std_logic;
+        -- Output Data
+        Out_Data        : out   std_logic_vector(Width_g-1 downto 0);
+        Out_Valid       : out   std_logic;
+        Out_Ready       : in    std_logic                                           := '1';
+        -- Configuration Ports (for RuntimeCfg_g = true)
+        Cfg_Period      : in    std_logic_vector(log2ceil(Period_g)-1 downto 0)     := toUslv(Period_g-1, log2ceil(Period_g));
+        Cfg_MaxSamples  : in    std_logic_vector(log2ceil(MaxSamples_g)-1 downto 0) := toUslv(MaxSamples_g-1, log2ceil(MaxSamples_g))
```
rbrglez, line 62: NITPITCK:

I wouldn’t set default values for the configuration ports. If RuntimeCfg_g = true, then the user should be required to explicitly provide values for the configuration ports. If they don’t want to do that, they should set RuntimeCfg_g = false instead. Providing default values for the configuration ports defeats the purpose of having them in the first place. What do you think?

src/base/vhdl/olo_base_rate_limit.vhd
```diff
@@ -0,0 +1,260 @@
...
+            variable SmoothLimit_v    : natural;
+        begin
+            -- Hold variables stable
+            v := r;
+
+            -- Register Runtime Configuration signals
+            if RuntimeCfg_g then
+                v.CfgPeriod     := to01(Cfg_Period);
+                v.CfgMaxSamples := to01(Cfg_MaxSamples);
+                -- to01 and resize required to workaroudn simulation issues. Functionally they do not
```
rbrglez: typo: workaroudn -> workaround

rbrglez (commented):
Again, the functionality of the code looks good overall. I just pointed out a few minor nitpicks.

By “nitpicks,” I mean small, non-blocking suggestions that you can safely ignore if you don’t agree with them.

src/base/vhdl/olo_base_rate_limit.vhd
```diff
@@ -0,0 +1,260 @@
...
+        signal In_Ready_i  : std_logic;
+        signal AllowSample : std_logic;
+
+    begin
+
+        -- Combinatorial process
+        p_comb : process (all) is
+            variable v                : TwoProcess_r;
+            variable OutputTransfer_v : boolean;
+            variable PeriodMin1_v     : natural;
```
rbrglez: Why is this variable named `PeriodMin1_v`? What does the `Min1` portion represent?

obruendl: I added a comment. I means "PeriodMinus1"

rbrglez (approved):

rbrglez:
@obruendl, looks good to me, you can merge it

Issue #239, opened 2025-11-05
Support CRC spanning multiple data beats in olo_base_crc_append / olo_base_crc_check


rbrglez:
I am implementing a generic ethernet datapath using AXI Stream. The datapath is N*8 bits wide and it includes byte-enable signals when wider than 8 bits. I want to use `olo_base_crc_append` and `olo_base_crc_check` to append and check the standard 32-bit Ethernet CRC, regardless of the data width.

Currently `olo_base_crc_append` and `olo_base_crc_check` assume that the CRC fits in a single data beat. They append or check only one data beat as the CRC, which is insufficient when the CRC spans multiple data beats (for example, a 32-bit CRC with an 8-bit data width requires 4 data beats).

It would be helpful to extend `olo_base_crc_append` to append CRCs that span multiple data beats and `olo_base_crc_check` to validate CRCs across multiple data beats. This would enable proper handling of standard-width CRCs across generic AXI Stream widths.

obruendl:
@rbrglez 

Extension to CRC wider than the data is an idea I had before. On the append side it should be easy but on the check side it is a bit less straightforward. I'll check but you should expect it to take some time.

Until then:

Are your packets always sized multiple of four bytes? It so, you could simply convert to 32 bit before the CRC and convert back after.

If this is not the case you can still already use the existing _olo_base_crc_ and build your own append/check logic around it.

I'll look into it. But I wanted to provide you already with workaro6nds I see to unblock you quickly.

rbrglez:
I’ve added a byte enable port (`In_Be`) to `olo_base_crc`, with minimal modifications.
Would it be possible to include this in the main `olo_base_crc` module?

I still need to decide how the unit should behave when `DataWidth_g` is not a multiple of 8. One option might be to disallow `In_Be` values that are not all `'1'` in such cases.

Here is entity declaration:
```vhdl
-- TODO: What if DataWidth not multiple of "Bytes"
-- TODO: What if In_Be = all 0
entity olo_base_crc is
    generic (
        DataWidth_g     : positive;
        Polynomial_g    : std_logic_vector;  -- according to https://crccalc.com/?crc=01&method=CRC-8&datatype=hex&outtype=bin
        InitialValue_g  : std_logic_vector := "0";
        BitOrder_g      : string           := "MSB_FIRST"; -- "MSB_FIRST" or "LSB_FIRST"
[...]

PR #240, opened 2025-11-10
IMPROVEMENT: Add backward-compatible Byte Enable support to olo_base_crc

rbrglez:
Initial commit of Byte Enable feature for `olo_base_crc`

Discussed in issue #239.

### DIFF
--- a/doc/base/olo_base_crc.md
+++ b/doc/base/olo_base_crc.md
@@ -170,4 +170,5 @@ Three packets are processed:
 **Note:** _olo_base_crc_ is configured with _ByteOrder_g_ = "MSB_FIRST".
 Even with this setting, _In_Be_ still follows the
 [**Trailing-Only Byte-Enable**](../Conventions.md#trailing-only-byte-enable) convention.
-On the last data beat, the byte-enable is not _In_Be_ = "10", but _In_Be_ = "01".
+On the last data beat, the byte-enable is not _In_Be_ = "10", but _In_Be_ = "01". With "MSB_FIRST"
+the most significant byte out of the enabled ones (according to _In_Be_) is processed first.


obruendl:
@rbrglez 

I have a pretty high level question before going really deep on the code: My main question is what is the use-case where we need to be able to allow *random byte enables* for *exactly this entity*?

## Random Byte Enables

I see a clear usecase for providing an option to ignore some bytes in the last word. Clearly, if we have a 32-bit wide data interface and there is a packet with 6 bytes, in the last word we have to disable the two highest bytes. But is there a real use-case for disabling radom bytes in the middle of a packet? This question is inspired by the definition of AMD for AXI4-Stream where they say TKEEP bits are only allowed to be de-asserted together with TLAST. I agree with AMD that this is the most common case - and if AMD has success with this definition, probably all other use-cases can be seen as minor. However, this might be wrong - so I'd like to know if you see a real scenario?

Regarding implementation, this would be much easier to handle and result in less extensive logic. It'd just require limiting the range of the loop for the last data-beat and it would not require alignment logic.

## Exactly this entity

I of course see the case where we know that the payload to build the CRC is somewhere burried in the frame (e.g. if we have 6 bytes header, we want to build the CRC for the payload from byte 7 to byte N). But I asked myself if it makes sense to integrate this into _olo_base_crc_. For me it would seem like a good feature providing a entity _olo_base_packet_extract_ that allows to properly remove a header of a given size (ideally defined through a port) and from there on extract a packet of a given size (in bytes, also defined by a port).

An option to not provide the size and just extract everything until the last byte (considering _Last_ and the _Be_) would be nice for cases where only a header exists but no trailer.

This entity could be used to extract the packet that goes into _olo_base_crc_ but also to do any other processing with it. Let's assume you get a UDP packet and you want to process the payload further. It would be nice if you can just say how much header you want to remove and you get the stripped (and properly aligned, so the first byte of the output stream is the first byte of the payload) at the output of that _olo_base_packet_extract_ entity.

If you have a real use-case for byte enable de-asserted in the middle, this entity could also do the compressing. Here I am not sure if this fits in or should be a separate "compress" one but I leave this descussion for the moment - let's first check whats your opinion on my thoughts in general.

## Conclusion

At a first glance the code looks sane - but I think we might be able to pull more value out of the alignment code if we properly separate the differnt functions. To me it feels like we loose lots of value if we add this only to one entity instead of offering it as general function.

rbrglez:
@obruendl , what about olo_base_crc Vunit testbench?

I really like Vunit testbenches for fixed point math components, which use cosimulation scripts to generate inputs and expected outputs.

This approach could be also used for olo_base_crc testbench.

What do you think? Should I invest some time for implementing this type of testbench?

rbrglez:
After re-reading your comment I see that you also don't want to "pack" / "compress" last data word.

> Regarding implementation, this would be much easier to handle and result in less extensive logic. It'd just require limiting the range of the loop for the last data-beat and it would not require alignment logic.

This is why I introduced the extractEnabledBytes() function. It assumes that In_Be takes valid patterns such as "1100", "0111",... and does not support invalid patterns like "1010", "1001", or "0110"

extractEnabledBytes() and packEnabledBytes() functions could be moved to olo_base_pkg_logic. I think that these functions would be useful elsewhere.

rbrglez:
Modified extractEnabledBytes() function to follow LSB first convention.

Moved extractEnabledBytes() and packEnabledBytes functions to olo_base_pkg_logic.

Added a simple vunit test for extractEnabledBytes. This test could be improved. I don't like how I had to add variable Expected2_v and Expected4_v, so that I could use check_equal() function without errors.

Should I remove function packEnabledBytes() from this pull request? Maybe add it in another pull request?

rbrglez:
@obruendl 

Could you take a look at the new testbench for olo_base_crc (olo_base_crc_be_tb)? I currently have two CRC modules running in parallel. With VUnit, I’m using a vc_stimuli component that generates the inputs, and I intended to add two response checkers to verify that the outputs from the CRC8 and CRC16 components are correct.

However, I’m running into a problem: when I enable both response-checking VC components at the same time, the testbench times out. Do you know how I could achieve this setup using the VUnit framework?

rbrglez:
@obruendl 

I have renamed the testbench olo_base_crc_be_tb to olo_base_crc_be_2byte_tb because the input data is hardcoded to 16 bits (2 bytes). Making the testbench generic to allow varying input sizes would increase complexity, making it harder to maintain and read.

> However, I’m running into a problem: when I enable both response-checking VC components at the same time, the testbench times out. Do you know how I could achieve this setup using the VUnit framework?

I have resolved this issue. Each olo_base_crc component now has its own VC stimulus generator and its own VC response checker. Additionally, I connected out_ready signal from the VC response checkers to the olo_base_crc components.

I could add more test cases and additional CRC algorithms to olo_base_crc_be_2byte_tb, but I decided not to, as I tend to go overboard with the size of testbenches. However, I can easily add more if you want.

rbrglez:
Added another testbench (olo_base_crc_be_4byte_tb), which tests longer be signal (4 bits)

src/base/vhdl/olo_base_pkg_logic.vhd
```diff
@@ -338,4 +346,70 @@ package body olo_base_pkg_logic is
         return getSetBitIndex(vec, fromMsb => false);
     end function;
 
+    function extractEnabledBytes (
+        data       : in std_logic_vector;
+        byteEnable : in std_logic_vector) return std_logic_vector is
+        variable ExtractedData_v : std_logic_vector(data'range) := (others => 'X');
+        variable High_v          : integer                      := byteEnable'low;
+        variable Low_v           : integer                      := byteEnable'low;
```
obruendl: This never gets assigned - hence it could be defined as constant.

src/base/vhdl/olo_base_crc.vhd
```diff
@@ -93,23 +95,32 @@ begin
...
+        variable InBit_v     : std_logic;
+        variable Out_v       : std_logic_vector(CrcWidth_c-1 downto 0);
+        variable InputHigh_v : natural;
     begin
         if rising_edge(Clk) then
-            -- Handle Input permutation (LFFSR always processes MSB first)
-            Input_v := In_Data;
+
+            if (BeMode_c and In_Last = '1') then
+                Input_v     := extractEnabledBytes(In_Data, In_Be);
```
obruendl: Why do we even need the `extractEnabledBytes()`? Can't we just assign all bits (the ones that are not enabled are then simply not interpreted)?

Of course we still should check if the byte enables are continuous and starting at the LSB.

src/base/vhdl/olo_base_pkg_logic.vhd
```diff
@@ -74,6 +74,14 @@ package olo_base_pkg_logic is
 
     function getTrailingSetBitIndex (vec : in std_logic_vector) return integer;
 
+    function extractEnabledBytes (
```
obruendl: I am not sure if we need this function - see comment in _olo_base_crc_.

src/base/vhdl/olo_base_pkg_logic.vhd
```diff
@@ -338,4 +346,70 @@ package body olo_base_pkg_logic is
...
+    begin
+        assert data'length mod 8 = 0
+            report "packEnabledBytes(): Data width must be a multiple of 8"
+            severity error;
+
+        assert byteEnable'length * 8 = data'length
+            report "packEnabledBytes(): Number of data bytes must match byteEnable length"
+            severity error;
+
+        if (byteEnable = onesVector(byteEnable'length)) then
```
obruendl: What is the reason for the if? Can't we just always execute the for loop - this would lead to simpler code - I'd not assume this to significantly change simulation runtimes and the like.

src/base/vhdl/olo_base_crc.vhd
```diff
@@ -119,6 +130,18 @@ begin
 
             -- Normal Operation
             if In_Valid = '1' and In_Ready_I = '1' then
+
+                -- Report a warning when In_Be is used improperly
+                if (BeMode_c and In_Last = '0') then
+                    assert In_Be = onesVector(In_Be'length)
+                        report "olo_base_crc: In_Be is ignored when In_Last is not asserted. All bytes are assumed enabled."
```
obruendl: The message should say what the detected problem is. currently it just describes the convention instead.
I'd suggest the mssage "In_Be de-asserted while In_Last='0', which is not supported" or the like

test/base/olo_base_crc/olo_base_crc_be_2byte_tb.vhd
obruendl, line 1: What is the reason to have 3 DUT instances and have code for all test-cases triplicated instead of having exactly one DUT and simply choose one out of 3 CRC settings based on a generic "CrcSettings_g : string" passed to the TB?

I believe this would shorten the TB code quite a bit - I thought I ask, if you like the idea you can change it, if you  do not agree and you want to keep it as is this also would be fine - but discussing things and deciding conciously always is the better option than not asking.

test/base/olo_base_crc/olo_base_crc_be_2byte_tb.vhd
```diff
@@ -0,0 +1,543 @@
...
+                Last_v := '1';
+                Be_v   := "11";
+                Data_v := x"0102";
+                push_axi_stream(net, AxisMaster_DectX_c,   Data_v, tuser => Be_v, tlast => Last_v);
+                push_axi_stream(net, AxisMaster_DvbS2_c,   Data_v, tuser => Be_v, tlast => Last_v);
+                push_axi_stream(net, AxisMaster_IsoHdlc_c, Data_v, tuser => Be_v, tlast => Last_v);
+
+                if (ByteOrder_g = "LSB_FIRST") then
+                    -- 0x0201
+                    check_axi_stream(net, AxisSlave_DvbS2_c,   x"C3",       msg => "CRC8/DVB-S2");
```
obruendl: Where did you get the expected response from? From the website linked in the documentation? If so - I think adding a comment stating this would be good. No need to repeat this everywhere - adding this comment in one place is sufficient.

test/base/olo_base_crc/olo_base_crc_be_2byte_tb.vhd
rbrglez: I started down this path to have multiple DUT's in the TB and I didn't even consider there is another way.

Now I have rewritten TB, so that I can pass CrcName_g AND DataWidth_g.

Thanks for bringing this up. Your question made me rethink the approach and TB is definitely better now because of it.

test/base/olo_base_crc/olo_base_crc_be_2byte_tb.vhd
rbrglez: I got the response from crccalc.com. Added comment in TB

src/base/vhdl/olo_base_pkg_logic.vhd
rbrglez: removed function

src/base/vhdl/olo_base_crc.vhd
rbrglez: Removed function.
I still need to add checking of In_Be input.

What should happen when In_Be isn't following LSB First convention or In_Be asserts aren't contiguous?

Is just assert enough? Maybe I should add another output port to olo_base_crc (Out_Error), Which is asserted when illegal value of In_Be is detected?

src/base/vhdl/olo_base_pkg_logic.vhd
rbrglez: Removed

src/base/vhdl/olo_base_pkg_logic.vhd
rbrglez: Removed packEnabledBytes function from this pull request.
I will add it in another pull request, not in this one, where focus should be only on the crc component

src/base/vhdl/olo_base_crc.vhd
rbrglez: done

rbrglez: Same issue as above: relying only on warnings/errors is not sufficient. Simulation will catch the problem, but on hardware the message will not be visible. We should consider adding an Out_Error port that asserts when In_Be violates the LSB-first convention or is de-asserted when In_Last = '0'.

rbrglez: After further reflection, I realized that you had already mentioned this earlier. olo_base_crc assumes that the incoming packet follows the LSB-first convention. If you are not sure whether the incoming data meets this requirement, you should use an additional component that “compresses” the stream. This component removes any deasserted byte-enables in the middle of the packet and ensures that the final byte-enable is correctly configured so the data adheres to the LSB-first format.

src/base/vhdl/olo_base_crc.vhd
obruendl: IMO asserts + proper documentation are enough.

doc/base/olo_base_crc.md
```diff
@@ -139,3 +140,27 @@ The following points are worth mentioning:
...
+
+### Byte enables
+
+The use of byte enables (_In_Be_) is limited to indicating that not all bytes are
+valid in the last beat of a packet. Any _In_Be_ being deasserted in words other
+than the last (_In_Last_ = '1') is not allowed and will result in simulation
+errors and undefined behavior in synthesis.
+
+For the last beat (_In_Last_ = '1'), if _In_Be_ is partially deasserted, it must
+follow the **LSB-first** convention: the least significant bit must always be
```
obruendl: I find the term _LSB-first_ convention a bit difficult - especially in an entity where we already have "LSB_FIRST" and "MSB_FIRST" for a generic.

What do you think of the name "Trailing-Only Byte-Enable" for this convention (to emphasize that we only disable training bytes)?

doc/base/olo_base_crc.md
```diff
@@ -139,3 +140,27 @@ The following points are worth mentioning:
 - _Out_Crc_ is updated exactly one clock cycle after the input data being transferred
 - The first packet shows backpressure (propagation of _Out_Ready_ to _In_Ready_)
 - The second packet is not terminated by _In_Last_ and the start of the third packet is indicated by _In_First_
+
+### Byte enables
```
obruendl, line 144: In the context of this entity I think you should clearly state that it is always the bytes starting at _In_Data[0]_ that must be enabled - _even if the crc is in MSB-First_ mode. 

The figure contains this case - but still I think emphasizing this would make sense.

doc/base/olo_base_crc.md
```diff
@@ -139,3 +140,27 @@ The following points are worth mentioning:
...
+asserted, and all asserted bits in _In_Be_ must be contiguous with no gaps.
+
+The example below shows a 16 bit input data bus. For simplicity, the
+signals _In_Ready_, _Out_Ready_ and _In_First_ are not shown, but they are fully
+supported in the implementation. The _olo_base_crc_ in this example uses "MSB_FIRST" byte order and
+the CRC-8/DVB-S2 algorithm.
+
+Three packets are processed:
+
+- The first contains 1 byte
```
obruendl: I'd suggest adding for each packet in the bullet-list the bytes that are processed (in the correct order).

doc/base/olo_base_crc.md
obruendl, line 144: Could you add a section about the "Trailing-Only Byte Enable" convention to "/doc/Conventions.md"?

src/base/vhdl/olo_base_crc.vhd
```diff
@@ -1,6 +1,6 @@
 ---------------------------------------------------------------------------------------------------
 -- Copyright (c) 2025 by Oliver Bruendler
--- Authors: Oliver Bruendler
+-- Authors: Oliver Bruendler, Rene Brglez
```
rbrglez, line 3: @obruendl , Is it ok that I added myself here?

src/base/vhdl/olo_base_crc.vhd
```diff
@@ -93,23 +98,37 @@ begin
...
-            -- Handle Input permutation (LFFSR always processes MSB first)
-            Input_v := In_Data;
-            if BitOrder_g = "MSB_FIRST" then
-                if ByteOrder_g = "LSB_FIRST" then
-                    Input_v := invertByteOrder(Input_v);
+
+            if (BeMode_c and In_Last = '1') then
+                BePlus_v                      := std_logic_vector(unsigned(In_Be) + 1);
+                assert (In_Be and BePlus_v) = zerosVector(In_Be'length)
+                    report "olo_base_crc: In_Be must have LSB asserted and all asserted bits must be contiguous. Trailing-Only Byte-Enable convention violated."
```
rbrglez, line 113: @obruendl  I have updated reports in this module. Review again

doc/base/olo_base_crc.md
rbrglez: done

doc/base/olo_base_crc.md
rbrglez: done

doc/base/olo_base_crc.md
rbrglez: done

doc/base/olo_base_crc.md
rbrglez: done

src/base/vhdl/olo_base_crc.vhd
obruendl: Sure

PR #243, opened 2025-12-03
BUGFIX: Make bytewise crc compatible with Quartus Prime Lite & Yosys

obruendl:
@rbrglez your implementation was not accepted by Quartus Prime Standard/Lite. The language support there really is horrible. I implemented a fix. Please quickly review.

Again I am happy to have a CI synthesis detecting such issues :-)

### DIFF
--- a/src/base/vhdl/olo_base_crc.vhd
+++ b/src/base/vhdl/olo_base_crc.vhd
@@ -66,13 +66,17 @@ end entity;
 architecture rtl of olo_base_crc is
 
     -- Constants
-    constant CrcWidth_c     : natural                                 := Polynomial_g'length;
-    constant BeMode_c       : boolean                                 := choose(DataWidth_g mod 8 = 0, true, false);
-    constant ZeroPoly_c     : std_logic_vector(CrcWidth_c-1 downto 0) := (others => '0');
-    constant InitialValue_c : std_logic_vector(CrcWidth_c-1 downto 0) := choose(InitialValue_g = "0", ZeroPoly_c, InitialValue_g);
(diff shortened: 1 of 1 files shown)


obruendl:
@rbrglez 

The code now runs with all tools except Yosys (cologne chip) - which seems to have issues with dynamic slicing.

Could you check if you find a workaround?

For installation of the OSS CAD Suite you can refer to the Cologne Chip homepage. To test, you can then run

```
cd ./tools/inference_test
python3 ./InferenceTest.py --entity olo_base_crc --yml ./yaml/base.yml --tool cologne
```

obruendl:
@rbrglez 

You probably can fix it by loops in the form:

```
for i in 0 to SomeConst_c loop
    if i <= SomeVar_v then
        a(i) <= b(i); -- you can also use ranges, like whole bytes here
    end if;
end loop;
```

This way the slices are constants and it is only a comparison against a variable. However, this may clutter the code. So:
- Place a comment to say why we are doing this thing
- Ideally do it in as many places as possible (if you need it often, probably extracting it into a function could help)

rbrglez:
> I updated your permissions. You should now be able to push to the crc-synth branch.

>I am not yet very used to GitHub collaborator permissions. Could you let me know if everything looks as intended? I would expect that you can push to the branch now but you should not be able to merge it and you should not be able to run GitHub actions manually. Can you quickly check and let me know?

I can push to branch and I don't think I can run github actions manually....at least I don't see the button to run the actions.

obruendl:
Funny enough, now Modelsim shoot us in the foot with a compile error ...

I fixed this one as well. Now it looks good.

The issues only emphasize how much the world requires a portable standard library for FPGA design :-)

Issue #159, opened 2025-06-01
SPI (or other interface) to AXI-Lite bridge


MiloradPetrovic1:
Greetings Oliver.

A few things.

**(1)**

Regarding the SPI to AXI-Lite bridge that is mentioned in the [feature list](https://github.com/open-logic/open-logic/wiki/Feature-Ideas).

Do you already have in mind how the messages from the SPI side should look like? Of course there needs to be a command + address + data, but for example, should there be a checksum for the transmitted SPI data? Also, are you aware of [this repo](https://github.com/airhdl/spi-to-axi-bridge)?

**(2)**

What are your thoughts on other **X to AXI-Lite** bridges? It would be nice to have a real, fast and easy to use memory mapped interface from a MCU/CPU to a FPGA.

Here are some options I think are interesting:
- QSPI to AXI-Lite
- FMC to AXI-Lite, FMC is a general purpose parallel bus used in STM32 microcontrollers.

**(3)**

Also, have you considered maybe **X to AXI-Stream** bridges?

- SPI to AXI-stream (much more easy to implement and use than the AXI-Lite variant)
- PSSI to AXI-stream, PSSI is a parallel interface used in STM32 microcontrollers and processors.
- uPP to  AXI-Lite, uPP is a parallel interface used in Texas Instruments C2000 microcontrollers.

obruendl:
Hi Milorad,

Thank you for the repo linked. I didn't know it - I added the linke to the point in the feature list, so I remember to check it and see if there even is a need to cover this in open-logic or if the existing repo is fine.

I did not think about the protocol in detail yet. Generally my mindset was to set-up a "byte-stream to AXI-Lite" protocol that could then be applied to UART as well.

For other busses (QSPI, FMC, and more) I'd not plan to invest any effort unless the community shows a real need and people start to ask for those things.

For the AXI-Stream bridges: The _olo_intf_spi_slave_ already has an AXI4-Stream interface, so no need to crate a separate bridge. And again, in general I'd not invest time unless a real need shows up.

Why do I care about UART/SPI to AXI-Lite but not so much about the other options:
- UART/SPI are by far the most commonly used ones and every microcontroller has them
- AXI-Lite is the must-have requirement to configure register banks in your FPGA from a microcontroller
- For significant data streaming and performant busses etween FPGA and CPU I currently see less use because very often if this need arises an SoC is the better solution anyways. But as mentioned, my decision here is not "not doing it" but "listening to the community".

[...]

PR #254, opened 2025-12-26
Byte Stream Bridge

rbrglez:
Created documentation for the byte_stream_bridge component discussed in issue #159

@obruendl, Can you review the documentation?

### DIFF
--- a/doc/axi/olo_axi_byte_stream_brige.md
+++ b/doc/axi/olo_axi_byte_stream_brige.md
@@ -1,16 +1,16 @@
 <img src="../Logo.png" alt="Logo" width="400">
 
-# olo_intf_byte_stream_bridge
+# olo_axi_byte_stream_bridge
 
 [Back to **Entity List**](../EntityList.md)
 
 ## Status Information
 
-![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/coverage/olo_intf_byte_stream_bridge.json?cacheSeconds=0)
-![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/branches/olo_intf_byte_stream_bridge.json?cacheSeconds=0)
-![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/issues/olo_intf_byte_stream_bridge.json?cacheSeconds=0)
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/coverage/olo_axi_byte_stream_bridge.json?cacheSeconds=0)
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/branches/olo_axi_byte_stream_bridge.json?cacheSeconds=0)
+![Endpoint Badge](https://img.shields.io/endpoint?url=https://storage.googleapis.com/open-logic-badges/issues/olo_axi_byte_stream_bridge.json?cacheSeconds=0)
 
-VHDL Source: [olo_intf_byte_stream_bridge](../../src/intf/vhdl/olo_intf_byte_stream_bridge.vhd)
+VHDL Source: [olo_axi_byte_stream_bridge](../../src/intf/vhdl/olo_axi_byte_stream_bridge.vhd)
 
 ## Description
 
--- a/doc/axi/olo_axi_byte_stream_brige.md
+++ b/doc/axi/olo_axi_byte_stream_brige.md
@@ -14,7 +14,18 @@ VHDL Source: [olo_axi_byte_stream_bridge](../../src/intf/vhdl/olo_axi_byte_strea
 
 ## Description
 
-### Overview
+This component implements the _Open Logic Byte Stream Protocol_ and acts as a
+bridge between a byte stream request/response interfaces and an AXI4-Lite
+master interface.
+
+It receives a Request Byte Stream through the _In Request Byte Stream
+Interface_. The incoming byte stream is decoded to determine the requested
+operation. Based on the decoded request, the component performs the
+corresponding read or write transaction on the AXI4-Lite master interface.
+
+Once the AXI4-Lite transaction completes, the component generates a Response
+Byte Stream containing the operation result and transmits it via the _Out
+Response Byte Stream Interface_.
 
 ## Interfaces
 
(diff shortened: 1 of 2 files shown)


doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+## Interfaces
+
+### Control
+
+| Name | In/Out | Length | Default | Description                                     |
+| :--- | :----- | :----- | ------- | :---------------------------------------------- |
+| Clk  | in     | 1      | -       | Clock                                           |
+| Rst  | in     | 1      | -       | Reset input (high-active, synchronous to _Clk_) |
+
+### In Request Byte Stream Interface
```
obruendl, line 39: I'd call this just *Request* or *Request Stream* interface.

Signals I'd name in the form *Request_Xxx* or *Rqst_Xxx* .

Reasoning: For the request *in* makes sense, but for the response this is somewhat misleading (because the response goes out). But calling the response *out* also does not make sense because logically the bytestream is one side an the axi the other. Hence I believe just naming them *Request* or *Rqst* and *Response* or *Resp* is somewhat more logical.

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+
+| Name            | In/Out | Length          | Default | Description                               |
+| :-------------- | :----- | :-------------- | ------- | :---------------------------------------- |
+| In_ReqReady     | out    | 1               | N/A     | AXI-S handshaking signal for _In_ReqData_ |
+| In_ReqValid     | in     | 1               | '1'     | AXI-S handshaking signal for _In_ReqData_ |
+| In_ReqData      | in     | 8               | -       | Input Byte Stream **REQUEST** Data        |
+
+### Out Response Byte Stream Interface
```
obruendl, line 47: I'd all this just *Response* or *Response Stream* interface. 

Signals I'd name in the form *Response_Xxx* or *Resp_Xxx*

Reasoning: See request interface

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+
+It receives a Request Byte Stream through the _In Request Byte Stream
+Interface_. The incoming byte stream is decoded to determine the requested
+operation. Based on the decoded request, the component performs the
+corresponding read or write transaction on the AXI4-Lite master interface.
+
+Once the AXI4-Lite transaction completes, the component generates a Response
+Byte Stream containing the operation result and transmits it via the _Out
+Response Byte Stream Interface_.
+
```
obruendl, line 29: Somewhere the documentation should mention what happens if the transaction times out (if there is no response on AXI). This can go here or in a section *Details*. It should just be covered somewhere. I don't think you have to put special effort into a specific behavior - documenting the natural behavior that results is fine. 

I'd also suggest to write down if the block accepts the next request before the last response is sent out or not. IMO it would make sense to add a (optional?) buffer for buffering a number of requests. Users might use faster interfaces than UART/SPI and in such a case the next request could arrive before the last one was executed.

I'd also add a note that the block is intended to bridge AXI4-Lite transactions through interfaces like UART and SPI to build FPGA-to-FPGA or PC-to-FPGA bridges.

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+| :--- | :----- | :----- | ------- | :---------------------------------------------- |
+| Clk  | in     | 1      | -       | Clock                                           |
+| Rst  | in     | 1      | -       | Reset input (high-active, synchronous to _Clk_) |
+
+### In Request Byte Stream Interface
+
+| Name            | In/Out | Length          | Default | Description                               |
+| :-------------- | :----- | :-------------- | ------- | :---------------------------------------- |
+| In_ReqReady     | out    | 1               | N/A     | AXI-S handshaking signal for _In_ReqData_ |
+| In_ReqValid     | in     | 1               | '1'     | AXI-S handshaking signal for _In_ReqData_ |
```
obruendl, line 44: I'd omit the default value here. I usually have it because many components could take an "always valid" input. But here I think this would not work out because the AXI4-Lite transactions can take time.

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+| M_AxiLite_... | *      | *      | *       | AXI4-Lite master interface. For the exact meaning of the signals, refer to the AXI4-Lite protocol specification. |
+
+## Open Logic Byte Stream Protocol
+
+The Open Logic Byte Stream Protocol is designed to operate over any
+byte-oriented transport medium, such as **SPI**, **I2C**, **UART**, or similar
+interfaces. The protocol is transport-agnostic.
+
+The protocol defines Request and Response frames, each identified by a
+distinct synchronization sequence.
```
obruendl, line 68: Add "... to support interfaces that provide raw byte-streams without any frame orientation" or the like

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+
+Every frame is protected by a CRC to detect transmission errors. If a CRC error
+is detected, the affected Request frame is discarded and the receiver resumes
+listening for the next valid Request frame.
```
obruendl, line 72: Add that no response is sent in this case.

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Request Frame Format
+
+**byte[1:0] - Synchronization Bytes**
+
+- Request Synchronization Bytes: **0x7A5A**
```
obruendl, line 78: Describe the sequence byte by byte (which one is first) so users cannot mininterpret endianness.

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+Reserved bits may be used in future revisions to extend data length encoding or to support AXI4 burst transactions.
+
+**byte[3] - Transaction ID**
+
+A monotonically increasing transaction ID assigned by the requester.
+
+The ID is echoed in the response and allows the requester to match responses to
+requests, enabling detection of lost or out-of-order frames.
+
+**byte[4:N-1] - Address and Data Bytes**
```
obruendl, line 105: Note the fields *Address* (always present)  *Data*  (present for writes only) separately.

I'd also suggest to use *N* for address bytes and *M* for data bytes and put the letterns N/M into the command byte table. Just to have the reference 100% clear. 

For both, address and data, describe which byte goes to which bits of the AXI-bus (that data[0] goes to tdata[7:0] etc.). Again, dodging endianness questions.

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+Frames with an invalid CRC are dropped.
+
+If bit errors corrupt the header, frame alignment may be lost. In this case,
+the receiver searches for the next valid synchronization sequence. If a sync
+sequence is detected within a corrupted frame, the CRC check will fail, and
+resynchronization continues.
+
+### Response Frame Format
+
+**byte[1:0] - Synchronization Bytes**
```
obruendl, line 128: Same here - describe byte by byte to avoid endianness questions.

doc/axi/olo_axi_byte_stream_brige.md
```diff
@@ -0,0 +1,159 @@
...
+
+- **Read response** : Contains data bytes returned from the AXI-Lite read
+- **Write response**: These Bytes are not present
+
+**byte[N:N+1] - CRC**
+
+A 16-bit CRC appended at the end of the frame.
+
+The CRC is calculated over the entire frame excluding the synchronization bytes
+
```
obruendl, line 158: I'd add an example of request/response for 2 byte address and 4 byte data, read as well as write transactions. Simple tables listing the exact bytes (one row per byte) and a textual description like "Write data 0xDEADBEEF to address 0x1234" should suffice.

Examples help users proofing their understanding and they serve as a good test case (e.g. if users operate the UART from a SW they wrote themselves)

obruendl (commented):
@rbrglez 

Documentation review done.

obruendl:
@rbrglez 

Do you have any updates on this one?

obruendl:
@rbrglez Are you still working on this one or shall I close the PR until you come back to the topic?

rbrglez:
you can close the PR.
I will reopen the PR when I'll have time

PR #255, opened 2026-01-06
FEATURE: Add olo_fix_bin_div (Binary Divider)

obruendl:


obruendl:
@rbrglez 

Could you give the documentation of _olo_fix_bin_div_ a brief review.

You are busy with the presentation, so it's OK if you do not go super-deep. It's mainly about checking if the text I wrote is understandable to somebody else than the author.

PR #256, opened 2026-01-19
Feature/olo base latency comp

obruendl:


obruendl:
@rbrglez 

Could you give the documentation of _olo_base_latency_comp_ a brief review.

You are busy with the presentation, so it's OK if you do not go super-deep. It's mainly about checking if the text I wrote is understandable to somebody else than the author.

PR #261, opened 2026-01-28
Add Riviera Pro option to Vunit simulation script

aruede:
This PR adds the `--rivierapro` option to the Vunit `run.py` script.
This has successfully been tested with Riviera Pro 2024.4 64-bit.


obruendl:
@aruede Do you know if I can get a free license anywhere? Ideally I'd have the possibility ti run simulations through rivierapro on my end for maintaining it.

aruede:
> @aruede Do you know if I can get a free license anywhere? Ideally I'd have the possibility ti run simulations through rivierapro on my end for maintaining it.

Unfortunately I don't think there are free licenses, except a 20-day evaluation license.

obruendl:
@aruede 

Did you already make progress?

I did not get any response from Aldec on a license. So currently I cannot support much.

However, I updated to the latest VUnit, probably you could check if this solves the issue.

obruendl:
@aruede 

Is there any update on this?

aruede:
It looks like the function `format_generic` in Vunits   Rivierapro backend only quotes values containing spaces. Values like `(1,8,4)` have no spaces, so they are passed unquoted. I think it then tries to interpret `(1,8,4)` as a VHDL aggregate expression rather than a string literal, fails, and falls back to the default generic value.

The fix would be to monkey-patch format_generic before vu.main() runs, adding quotes around values containing (, ), or , when using Rivierapro, like this:

```
if os.environ.get('VUNIT_SIMULATOR') == 'rivierapro':
    import vunit.sim_if.rivierapro as _rp
    def _rivierapro_format_generic(value):
        value_str = str(value)
        if any(c in value_str for c in '(), '):
            return f'"{value_str}"'
        return value_str
    _rp.format_generic = _rivierapro_format_generic
```

I will let the whole simulation run and paste the results.

aruede:
The fix (see last commit) has worked, all tests pass.
image
Simulation xml:
[sim.xml](https://github.com/user-attachments/files/26970629/sim.xml)

obruendl:
@aruede 

I added the documentation entry plus a warning when running the simulation to clarify that I currently cannot actively maintain riviera pro due to unavailability of a license. Just to be transparent to other users.

With that the situation is clear for everyone and I happily integrate your work. Thank you for fixing the issue.

PR #283, opened 2026-04-05
FEATURE: Add olo_fix_sample_hold

obruendl:


obruendl (commented):
review done

rbrglez:
The olo_fix_sample_hold module could be a useful component even for applications that don’t rely on fixed-point dependencies.

Would it be possible to introduce a more generic olo_base_sample_hold module with no fixed-point dependencies? Then, olo_fix_sample_hold could be implemented as a wrapper around this base module, adding the necessary fixed-point dependencies on top.

obruendl:
@rbrglez 

I had exactly the same idea originally but came to the conclusion that it is too trivial in an RTL context and the only value is to stay on structural level for closed loop fixed-point cases.

You comment shows that potentially my conclusion was wrong. It's no significant effort anyways, so I'll add _olo_base_sample_hold_ and convet _olo_fix_sample_hold_ to a wraper around it.

obruendl:
@rbrglez 

I have added _olo_base_sample_hold_ 

Please check and provide feedback

Issue #284, opened 2026-04-08
Low performance on olo_base_fifo_packet for maximum packet size


tasgomes:
The Packet FIFO implemented in `olo_base_fifo_packet.vhd` works a bit differently of what I would expect from a Packet FIFO but it does offer other features that made me use it in a project.

The first expectation is that I would expect the following generics from a Packet FIFO:
- One to define the size of a packet
- One to define the number of packets

With this you end up with a Packet FIFO of size of `<number_of_packets> x <size_of_packet>`. The disadvantage is the waste of memory for smaller packets.

Instead, this Packet FIFO offers the following generics:
- One to set the number of FIFO entries (that also defines the maximum size of a packet for the packet drop functionality)
- One to define the number of packets

Note 1: the number of FIFO entries (or maximum size of a packet) is also used by this FIFO to drop packets above the maximum size.
Note 2: it can store multiple packets as long as they fit inside the number of FIFO entries.

With this you end up with a Packet FIFO of size of `<number_of_entries>`.

Fine for now. 

The hidden problem (at least not clearly visible at first) is that this Packet FIFO will degrade the throughput of the system when packets of maximum size are being transmitted along the system. When a packet is stored in the Packet FIFO, the signal `In_Ready` at the input interface will be kept low until the whole packet is read on the output interface.

A practical example where the maximum size of a packet is set to 16 KB. When 8 KB packets flow through this FIFO, we get ~100% throughput (see left side of the picture below). When 16 KB packets flow though this FIFO we get ~50% throughput (see right side):

<img width="1295" height="237" alt="Image" src="https://github.com/user-attachments/assets/46c2790a-49f1-4576-a30f-75ae5ff056c5" />

The conclusion is that this block lowers the performance/throughput for the scenario where maximum size of packet is used. Since usually the bigger the packet size the better the performance, specially for DMA transactions, this is a bit of a bummer.

With a traditional `<number_of_packets> x <size_of_packet>` design and `number_of_packets ≥ 2`, the write side can fill one buffer while the read side drains another, achieving 100% throughput even at maximum packet size. The shared-buffer approach in `olo_base_fifo_packet` makes this impossible when packets consume the full depth.

obruendl:
@tasgomes 

I 100% agree on the analysis. 

Question from my end: what shall be fixed? Is it just about outlining the behavior in the docimentation? If so, could you open a PR with the extended documentation?

[...]

PR #286, opened 2026-04-10
Merge from feature/fifo-packet-improvements

tasgomes:
@obruendl 

I have implemented a specific testbench to check the throughput of the Packet FIFO. This testbench constantly monitors the input TREADY of the Packet FIFO, ensuring it is never de-asserted to allow 100% throughput.

I have also implemented the changes we discussed in https://github.com/open-logic/open-logic/issues/284 to the RTL code.

However, I think I found another issue that was already present in the Packet FIFO before this feature. In order to demonstrate that, I generated different test cases for this new testbench.

```
olo_tb.olo_base_fifo_packet_perf_tb.MaxPackets_g=2-Depth_g=32-MaxPacketSize_g=16.test_throughput
olo_tb.olo_base_fifo_packet_perf_tb.MaxPackets_g=2-Depth_g=32-MaxPacketSize_g=-1.test_throughput
olo_tb.olo_base_fifo_packet_perf_tb.MaxPackets_g=8-Depth_g=32-MaxPacketSize_g=16.test_throughput
olo_tb.olo_base_fifo_packet_perf_tb.MaxPackets_g=8-Depth_g=32-MaxPacketSize_g=-1.test_throughput
```

The test cases for legacy behavior set `MaxPacketSize_g=-1`. Note how the same test passes when `MaxPackets_g=8`  but fails when `MaxPackets_g=2`:

image

I have noticed that all your existent testbenches use hardcoded `MaxPackets_g=4` so this might be unnoticed until now.

When `MaxPackets_g=8`, we can see that input TREADY is never de-asserted for the duration of a clock cycle when packets of different size flow through the FIFO:

image

However, when `MaxPackets_g=2`, the input TREADY is de-asserted after the reception of one packet (even though there should be room for 2 packets since `MaxPackets_g=2`:

image

The `Depth_g` of the FIFO did not change between test cases, hence there should be room to store the data in there at 100% throughput.

Could you please have a look into this? I guess the problem comes from the logic that handles packets.

I am not yet sure but this could also be the reason why the scenario I wanted to test still fails:

```
python ./sim/run.py --modelsim olo_tb.olo_base_fifo_packet_perf_tb.MaxPackets_g=8-Depth_g=32-MaxPacketSize_g=16.test_throughput
```

image

Note how the first two larger packets can be written without back-pressure (better than the legacy behavior) but then eventually TREADY lowers again.

Feel free to push into this branch.


obruendl:
@tasgomes 

Regarding _In_Ready_ going low when the buffer only has space for two packets: Without looking much at the code, this seems to make sense to me. We only can free the buffer after the packet was consumed entirely (because the user could request repeating the packet until this point in time). So it makes sense that we cannot accept packets back-to-back when they are exactly half of the buffer size. Do you agree?

Would it make sense to optimie the "DROP_ONLY" implementation of the FIFO? There we could actually free up space in the FIFO as soon as a sample is consumed on the output (because users cannot ask for repeating a packet). So here we could achieve full throughput easily without running into timing issues. Would this be sufficient from your perspective?

tasgomes:
@obruendl hum. In my case, I don't use the Repeating (`Out_Repeat`) feature but I do use the Skip feature (`Out_Next`). This is why I use the `FULL` configuration and I have the feeling that in this scenario I deserve the full throughput 😄 

For instance, if you want to still be able to drop a packet on the output side because a CRC failed or some other error was detected in the packet header, then the use of `Out_Next` to skip the current packet is still very useful.

obruendl:
@tasgomes 

I checked. At least the fix would not be simple. The behavior actually is clearly documented

> Note: Due to implementation reasons the FIFO introduces one stall cycle per packet on the read-side. Hence the FIFO throughput is suboptimal for very small packets.

So on the read-side one cycle per packet is lost. This eventually translates back to the write side.

This is a fundamental property of how the read FSM works. The only way around I'd see is to introduce a "DROP-SKIP-ONLY" mode and change the behavior completely (at least on the read side). However, this will be non-straight-forward and create some long combinatorial paths on the read side I suspect.

tasgomes:
@obruendl I understand.

Since I need to use the `Out_Next` signal, and therefore the "FULL" configuration, I think I have to live with this current setup for now.

However, simply having the possibility to set the FIFO size and the maximum packet size separately allows me to go from 50% throughput to something above 90%. I haven't measured it on HW yet, but I can clearly see this gain in simulation.

Would it make sense to proceed with this change as-is, and apply extra enhancements in a future PR?

obruendl:
@tasgomes 

Yes, this would make sense.

tasgomes:
@obruendl for the 50% throughput the configuration is 128-bit width and a depth of 16 KB. For the >90% throughput setup, it would be 128-bit width, a depth of 32 KB and a maximum packet size of 16 KB.

I can't recall for sure, but I think have tried that and there is an assertion that forces the depth to be power of 2.

obruendl:
@tasgomes 

You costed me my evening ;-)

Check-out the implementation on the branch _feature/fifo-packet-improvements_ in the open-logic main repo.

It has a new parameter "Optimization_g". Set it to "throughput" and use the new "FeatureSet_g=drop_skip_only".

The code is not fully cleaned yet but it seems to work. It should allow 100% throughput for packets bigger than 3 beats and at the same time you can skip.

Please review and test thorroughly.

If you find it fine - it'd be great if you could compare the maximum achievable clock speed with "Optimization_g=throughput" ves "Optimization_g=speed". I am interested to know if there is any difference or if we can aim for "Optimization_g=speed" always.

EDIT: Looks likecI pushed to your fork instead of mine. Sorry fir that...

obruendl:
@tasgomes 

I cleaned up code and documentation. Please review.

Especially please review the documentation to check if the updates are clear and there are no inconsistencies. 

I hope you are happy with the levels of support you got on Open Logic ;-)

tasgomes:
Thanks @obruendl. I have to find time to review and test this under the project where I am using this block. I will let you know when I do so.

tasgomes:
@obruendl I ran the simulations as this requires low effort and provides fast feedback. 

What I did:

- Run regression with "Optimization_g=throughput" and "FeatureSet_g=drop_skip_only" for my project 
  - The drop and skip functionality seems to work as before ✅ 
  - The throughput test now passes ✅ 
  - A Packet Size test case fails 🔴 
   
- Run regression with "Optimization_g=speed" and "FeatureSet_g=drop_skip_only" for my project 
  - The drop and skip functionality seems to work as before ✅ 
  - The throughput test still fails 🔴 
  - A Packet Size test case passes ✅ 

### Issue 1
  
With this sentence:
  
>  I am interested to know if there is any difference or if we can aim for "Optimization_g=speed" always.
  
I assume you were expecting also full throughput when "Optimization_g=speed"?

I set "Optimization_g=speed" in the `olo_base_fifo_packet_perf_tb` and it also fails, so if you expect it to work you can reproduce it already there. Maybe we should generate an extra test case by setting `Optimization_g` from the `run.py`.

### Issue 2

Why the Packet Size test fails?

The reason this test is failing is because I have another `olo_base_fifo_sync` running in parallel with the Packet FIFO to store the value of `TKEEP` so I can compute the correct size of the packet (in bytes) on the output of the Packet FIFO. 

For that the generic `olo_base_fifo_sync.Depth_g` was set to `olo_base_fifo_packet.MaxPackets_g`. With that I can use `olo_base_fifo_packet.In_Valid`, `olo_base_fifo_packet.In_Last` and `olo_base_fifo_packet.In_Ready` to control `olo_base_fifo_sync.In_Valid` without caring for `olo_base_fifo_sync.In_Ready`  since there should be always enough room on the `olo_base_fifo_sync` FIFO.

It seems the `olo_base_fifo_sync` has a delay of one clock cycle when a word is read until it propagates to `In_Ready` which creates an issue.

image

So this looks more like an issue of the `olo_base_fifo_sync` that prevents us from coupling it with the `olo_base_fifo_packet` so we can follow up with that in another issue. For the moment, as workaround, I increase the `olo_base_fifo_sync.Depth_g` by one.

obruendl:
@tasgomes 

## Issue 1

I have update the _olo_base_fifo_packet_perf_tb_ here in the PR a bit because it produced false positives. I think I understood the changes required but I need your review on them.

When you say the _olo_base_fifo_packet_perf_tb_ fails, did you point to your local copy or the one in this branch?

## Issue 2

Yes, there is a latency of one clock cycle - I recently did I presentation slide about the exact behavior, that's why I know by heart.

This issue then I understand as an issue in your application (how you wire up the two FIFOs) and not within the _olo_base_fifo_packet_ and you could fix it on your end. CorrecT?

tasgomes:
### Issue 1

It fails using this commit:
```
Commit: d8bdb801ee008d6a6c84d6e0521f0907c62bb19e [d8bdb80]
Parents: e711ce76b9
Author: Oliver Bründler <oli.bruendler@gmx.ch>
Date: Wednesday, 22 April 2026 00:19:39
Committer: Oliver Bründler
```

and this local change:

image

Maybe this is normal, this is why I need to know if you expect the "speed" option to also have 100% In_Ready or not. If you don't expect it, then it should be fine and Issue 1 can be closed.

### Issue 2

Well, I am not sure, because the issue only appears when I use `olo_base_fifo_packet.Optimization_g = "SPEED"`  my test case passes because the `olo_base_fifo_packet` seems to have the same latency of a at least one clock cycle. If I use `olo_base_fifo_packet.Optimization_g = "THROUGHPUT"` then my test case fails. It looks like the latency changes (because now `olo_base_fifo_packet.In_Ready` is always ready while `olo_base_fifo_sync.In_Ready` is not... so it is like I can't pair those two blocks anymore for the throughput option.

I do have a work-around which is increasing the `olo_base_fifo_sync.Depth_g`.

obruendl:
@tasgomes 

SPEED is the same behavior as before the change" THROUGHPUT removes the stsll cycle at the cost of lower clock frequency. See documentation (also already updated) 

If SPEED would not have any drawbach, I'd not provide a generic.

Hence your two issues seem logical:

- throughput simulation with SPEED fails because this is the same logic as before the change.
- In_Ready contains stall cycles for SPEED for the same reason

tasgomes:
@obruendl in that case all my tests pass with the new throughput mode. Seems like you solved my issue.

I will try to review code and documentation as soon as possible.

doc/base/olo_base_fifo_packet.md
```diff
@@ -52,30 +52,45 @@ for different technologies (some technologies implement one, some the other beha
...
+
+The FIFO can be optimized for two targets.
+
+- _SPEED_: The FIFO is optimized for best clock speed.
+  - Due to implementation reasons the FIFO introduces one stall cycle per packet on the read-side in this mode. Hence
+    the FIFO throughput is suboptimal for small packets.
+  - This mode is supported for all _FeatureSet_g_ options.
+- _THROUGHPUT_: The FIFO is optimized for best throughput.
+  - In this mode no stall cycles are introduced on the read side. Hence the FIFO can achieve a throughput of 100% even
+    for small packets.
```
tasgomes: Please confirm but I think we could be more clear and refer to packets above 3 beats.

tasgomes (commented):
@obruendl everything looks good to me. There is only one point that could be more clear.

doc/base/olo_base_fifo_packet.md
obruendl: Added a note.

I believe it is actually the overall latency of the block that plays a role. For a FIFO containing more than 2 packets, even smaller packets will perform well. But I think going too deep in documentation only confuses users. The 3 beats is what the TB checks, so it's safe to document this - if the FIFO performs better for some cases that's still allowed.

PR #294, opened 2026-04-24
FEATURE: Add olo_ft area with ECC-protected single-port RAM

rustyqt:
Introduces a new functional area `ft` (fault tolerance) for radiation-sensitive
  designs (space, avionics, automotive, high-altitude). All entities use SECDED
  Hamming code (Single Error Correction, Double Error Detection) and follow a
  transparent wrapper pattern: ECC is invisible to the user, with data encoded on
  write and decoded/corrected on read.

  This PR is the first in a series and contains the foundational pieces only:

    src/ft/vhdl/olo_ft_pkg_ecc.vhd
      SECDED Hamming code package (eccEncode / eccCorrectData /
      eccSyndromeAndParity / eccSecError / eccDedError). Overhead is 5 parity
      bits for 8-bit data, 6 for 16-bit, 7 for 32-bit, 8 for 64-bit, 9 for
      128-bit.

    src/ft/vhdl/olo_ft_ram_sp.vhd
      ECC-protected single-port RAM. Wraps olo_base_ram_sp with a wider internal
      word to store the codeword. Exposes the same interface as olo_base_ram_sp
      plus:
        - WrEccBitFlip (2 bits) for error injection / BIST
        - RdSecErr / RdDedErr status flags - EccPipeline_g generic for breaking the ECC-decode critical path

  Testbench:
    test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd — VUnit testbench covering basic read/write, SEC injection + correction, DED injection + detection, and multi-address isolation. 40 configurations swept across RamBehavior_g (RBW, WBR), RdLatency_g (1, 2), Width_g (8, 16, 32, 64), and EccPipeline_g (0, 1).

  Documentation:
    doc/ft/olo_ft_pkg_ecc.md and doc/ft/olo_ft_ram_sp.md. doc/EntityList.md gains a new "ft" section.

  Build / tooling integration:
    - sim/run.py registers the new olo_ft test-config module
    - sim/test_configs/olo_ft.py declares the testbench parameter sweeps
    - tools/libero/import_sources.tcl includes src/ft/vhdl in the import loop
    - tools/inference_test/yaml/ft.yml adds synthesis inference configurations for olo_ft_ram_sp across width / latency / pipeline combinations
    - compile_order.txt regenerated to include the new ft entries

  Also bundles a small tooling fix in sim/run.py: the nvc.global_flags option is
  not recognized by older VUnit versions, so the call is now wrapped in
  try/except ValueError. Without this, sim/run.py fails to load on those
  installations.

**Open for Discussion:**
 - **WrEccBitFlip bit width**:  It could be useful for verification purposes to increase the bit width to total stored bits ( data width + parity bits). This way any individual bit could be flipped to fully verify the SECDEC codec.
 -  **RdSecErr** and **RdDedErr** naming: Single-error-correction is no error condition so the naming might be misleading, could be renamed to RdSecWarn. Or something completely different, e.g.: **RdEccSec** and **RdEccDed**.

**Verification State:** The 40 testbench configurations pass locally with GHDL.

**Disclosure on GenAI Usage:** This PR was prepared with assistance from Claude Code (Anthropic).


src/ft/vhdl/olo_ft_pkg_ecc.vhd
```diff
@@ -0,0 +1,251 @@
...
+
+end package;
+
+---------------------------------------------------------------------------------------------------
+-- Package Body
+---------------------------------------------------------------------------------------------------
+package body olo_ft_pkg_ecc is
+
+    -- Helper: check if a positive integer is a power of 2
+    function isPow2 (
```
obruendl: Use _isPower2()_ from _olo_base_pkg_math_ inestead

src/ft/vhdl/olo_ft_pkg_ecc.vhd
```diff
@@ -0,0 +1,251 @@
...
+-- Libraries
+---------------------------------------------------------------------------------------------------
+library ieee;
+    use ieee.std_logic_1164.all;
+    use ieee.numeric_std.all;
+
+---------------------------------------------------------------------------------------------------
+-- Package Header
+---------------------------------------------------------------------------------------------------
+package olo_ft_pkg_ecc is
```
obruendl, line 32: Please create a TB for the _olo_ft_pkg_ecc_ as well.

First, I do have one TB for everyp roduction code file. This has proven useful, also for debugging and unerstanding code.

Second, it allows you to cover things that the usage of the package in entities does not. For example in this case, you can test in the package TB if eccCorrectData really can correct a bit-flip at any position even if the error injection in the RAM only can flip two positions.

src/ft/vhdl/olo_ft_pkg_ecc.vhd
```diff
@@ -0,0 +1,251 @@
...
+        DataWidth : positive
+    ) return positive;
+
+    -- Calculate total codeword width (data + Hamming parity + overall parity)
+    function eccCodewordWidth (
+        DataWidth : positive
+    ) return positive;
+
+    -- Encode data into a SECDED codeword
+    function eccEncode (
```
obruendl, line 43: Can you also provide two entities _olo_ft_ecc_encode_ and _olo_ft_ecc_decode_:

* _olo_ft_ecc_encode_ 
  * Encode a data word (add ECC)
  * Optionally registered
  * Also add the option for error injection directly here.
* _olo_ft_ecc_decode_:
  * Decode ECC word
  * Correct single flips, detec double flips
  * Optionall registered

Why do I ask to create entities although all the functionality is available in the package?

For verilog users functions from VHDL packages are not accessible. Entities are. I commit to make functionality available to veriloggers as well - and ECC is a core functionality of the new _olo_ft_ area. I see use-cases beyond RAM protection - one could apply ECC for FEC when transmitting data over a link or on an AXI4-Stream that is buffered in external DDR memory (which might not be ECC protected - sometimes the HW is given).

src/ft/vhdl/olo_ft_ram_sp.vhd
```diff
@@ -0,0 +1,140 @@
...
+                EccDedPipe(1) <= Rd_EccDedI;
+                DataPipe(2 to EccPipeline_g)   <= DataPipe(1 to EccPipeline_g - 1);
+                EccSecPipe(2 to EccPipeline_g) <= EccSecPipe(1 to EccPipeline_g - 1);
+                EccDedPipe(2 to EccPipeline_g) <= EccDedPipe(1 to EccPipeline_g - 1);
+            end if;
+        end process;
+        RdData   <= DataPipe(EccPipeline_g);
+        RdEccSec <= EccSecPipe(EccPipeline_g);
+        RdEccDed <= EccDedPipe(EccPipeline_g);
+    end generate;
```
obruendl: I suggested in the package that ECC encode and decode are implemented in entities.

The entity would also contain optional pipelieing. Hence the whole code-block here would be prelaced by the ECC decode entity.

We can see here also a benefit to the _olo_ft_ area without even thinking about verilog use-cases: If we implement this as an entity, we do not have to replicate this whole code-block in all the different RAMs but we just instantiate the one entity.

src/ft/vhdl/olo_ft_ram_sp.vhd
```diff
@@ -0,0 +1,140 @@
...
+    use work.olo_ft_pkg_ecc.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity
+---------------------------------------------------------------------------------------------------
+entity olo_ft_ram_sp is
+    generic (
+        Depth_g       : positive;
+        Width_g       : positive;
+        RdLatency_g   : positive := 1;
```
obruendl: Rename to "RamRdLatency_g" because this is the part excluding ECC. Just to make it clear to users.

src/ft/vhdl/olo_ft_ram_sp.vhd
```diff
@@ -0,0 +1,140 @@
...
+
+library work;
+    use work.olo_base_pkg_math.all;
+    use work.olo_base_pkg_attribute.all;
+    use work.olo_ft_pkg_ecc.all;
+
+---------------------------------------------------------------------------------------------------
+-- Entity
+---------------------------------------------------------------------------------------------------
+entity olo_ft_ram_sp is
```
obruendl, line 34: ## Scrubbing Concept

This comment is **NOT** meant as "change request" but to start a discussion early that is targeted at the next PR you gonna open.

I have some ideas/questions to the scrubbing concept. I number them (this helps keeping the answers organized, you can smply refer to numbers).

1. When we add scrubbing, could we add it to _olo_ft_ram_sp_ instead of creating a new entity? From a user perspective I think it would be great having an _olo_ft_ram_sp_ that simply has a generic to select if scrubbing is needed or not.
2. For scrubbing, I see different implementation options. Do you implement all of them or only one?
    * Scrubbing has priority. When it runs the user cannot access the RAM.
        * In this case, I see the option that scrubbing checks one cell every N clock cycles (evenly distributed) or the option that it runs in bursts (it checks all cells but does so only every M clock cycles).
        * This option is interfering with the user - but it gives guarantees how often scrubbing runs
    * Scrubbing only uses cycles the user does not use the RAM. It automatically detects them and uses them to execute the scrubbing 
        * This option does not interfer with the user, but it does not guarantee that scrubbing is executed in a given timeframe
     * The two could be combined (scrubbing uses stall cycles only but if it could not walk the full RAM within M clock cycles, it takes priority over the user access to guarantee a certain scrubbing rate)
     * For the single port RAM specifically - there would be the option to use a TDP RAM and give one port to the user and use the other for scrubbing.
     * I am definetly not expert in this area, so I am not sure what is common. But these are the options I see and it could be that we want more than one of them (selectable through a generic)
3. Because we gonna add scrubbing to different RAM types, would it make sense to implement scrubbing as a separate entity (it can be a "private" one, not intended for the end-user to use and not fully documented) and simply instantiate this entity in all RAMs with scrubbing? Again, beyond avoiding code duplication this also could help with easy test-coverage for corner cases.

doc/ft/olo_ft_pkg_ecc.md
```diff
@@ -0,0 +1,24 @@
...
+
+# olo_ft_pkg_ecc
+
+[Back to **Entity List**](../EntityList.md)
+
+## Status Information
+
+VHDL Source: [olo_ft_pkg_ecc](../../src/ft/vhdl/olo_ft_pkg_ecc.vhd)
+
+## Description
```
obruendl, line 15: In the package, please document every procedure separately. No super long description needed - but I'd like to keep up the idea that for every procedure there is documentation.

doc/ft/olo_ft_ram_sp.md
```diff
@@ -0,0 +1,95 @@
...
+
+The ECC encoding is combinational on the write path. The internal RAM (an instance of _olo_base_ram_sp_ with wider
+word) provides the configurable read pipeline (_RdLatency_g_). The ECC decoding is combinational after the read
+pipeline, so the error flags are time-aligned with the read data.
+
+When _EccPipeline_g_ > 0, additional register stages are inserted after the ECC decode logic. This breaks the
+combinational path between the RAM output and the corrected data output, which can help close timing at high clock
+frequencies. The total read latency becomes _RdLatency_g_ + _EccPipeline_g_ clock cycles.
+
+### ECC Overhead
```
obruendl: I think we will have quite some some documentation content that is not entity specific but applies to the FT area in parallel.

ECC overhead, error injection, scrubbing considerations etc. will apply to several RAMs. I could imagine there is other content that is applicable to multiple entities as well.

What I normally do - and I'd also suggest it here - is to write a _olo_ft_principles.md_ file documenting those core principles in one place and the documentation of individual entities can then refer to this file. This way we avoid duplication of the documentation.

You can for example check [clock_crossing_principles](https://github.com/rustyqt/open-logic/blob/feature/fault-tolerant/doc/base/clock_crossing_principles.md) or [olo_fix_principles](https://github.com/rustyqt/open-logic/blob/feature/fault-tolerant/doc/fix/olo_fix_principles.md)

src/ft/vhdl/olo_ft_ram_sp.vhd
```diff
@@ -0,0 +1,140 @@
...
+        RamBehavior_g : string   := "RBW";
+        EccPipeline_g : natural  := 0
+    );
+    port (
+        Clk          : in    std_logic;
+        Addr         : in    std_logic_vector(log2ceil(Depth_g) - 1 downto 0);
+        WrEna        : in    std_logic                               := '1';
+        WrData       : in    std_logic_vector(Width_g - 1 downto 0);
+        WrEccBitFlip : in    std_logic_vector(eccCodewordWidth(Width_g) - 1 downto 0) := (others => '0');
+        RdData       : out   std_logic_vector(Width_g - 1 downto 0);
```
obruendl: For a normal RAM that in most cases is used with _RdLatency_g=1_ having a _RdDataValid_ port is not really required. But TBH - I could consider adding it as well.

However, here we gonna have more delay (at least when _EccPipeline_g>0_. IMO it could make sense adding this _RdDataValid_ or _RdValid_ port. It could simplify user logic around the RAM I believe and the  effort for adding it is marginal.

doc/ft/olo_ft_ram_sp.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Interfaces
+
+| Name         | In/Out | Length                | Default | Description                                                  |
+| :----------- | :----- | :-------------------- | ------- | :----------------------------------------------------------- |
+| Clk          | in     | 1                     | -       | Clock                                                        |
+| Addr         | in     | _ceil(log2(Depth_g))_ | -       | Address                                                      |
+| WrEna        | in     | 1                     | '1'     | Write enable                                                 |
+| WrData       | in     | _Width_g_             | -       | Write data                                                   |
+| WrEccBitFlip | in     | _eccCodewordWidth(Width_g)_ | (others => '0') | ECC error injection for testing/BIST. Each '1' bit XORs (flips) the corresponding bit of the stored codeword. Popcount 1 = SEC-correctable, popcount 2 = DED-detectable.<br>See [Error Injection](#error-injection). |
```
obruendl: _eccCodeWordWidth()_ is a non-trivial one. Please add a reference to the meaning of the function (either describe it in the package and reference to there or describe it in the _olo_ft_principles_ and add a reference to there)

Detail: I usually write "all 0" instead of "(others => '0')" - just to make th edocumentation language independent. I don't want veriloggers to feel open logic is meant to be used from VHDL only - it clearly can be used from both languages.

doc/ft/olo_ft_ram_sp.md
```diff
@@ -0,0 +1,95 @@
...
+| RdData       | out    | _Width_g_             | N/A     | Read data (corrected if a single-bit error was detected)     |
+| RdEccSec     | out    | 1                     | N/A     | Single error corrected flag. '1' when a single-bit error was detected and corrected. |
+| RdEccDed     | out    | 1                     | N/A     | Double error detected flag. '1' when an uncorrectable double-bit error was detected. Read data is unreliable in this case. |
+
+## Detailed Description
+
+### Architecture
+
+```
+Write path:  WrData -> eccEncode -> XOR bit-flip injection -> wider internal RAM
```
obruendl: Add a figure. A small block diagram always helps user understanding what goes on.

doc/ft/olo_ft_ram_sp.md
obruendl: Currently _WrEccBitFlip_ must be valid when _WrEna='1'_. A logical choice.

However, I asked myself if it would be useful for verification and error injection on HW, to make two ports (_ErrInj_BitFlip_ and _ErrInj_Valid_). The idea would be that the bit-flip is latched when _ErrInj_Valid='1'_ and applied directly (as up to now) if _WrEna='1'_ in the same clock cycle. But when no data is written in this clock cycle, the bit-flip would stay stored and be applied to the next data word written.

This would allow to inject errors without having to know when exactly data is written. It would allow clearly separating the test logic from the actual datapath. This IMO could be useful for clean design of error injection logic.

Additionally it could be helpful for simulations. Your top-level TB may want to operate the error injection ports of a RAM deep in the design. At this point it would be nicer if the top-level TB does not have to care in which exact clock cycle data is written to the RAM.

obruendl (commented):
@rustyqt 

Pretty solid work I'd say! Thumbs up for this. I seldom got pull requests where I could dive so directly into the actual business logic without having to explain many "this is how it is done in open logic" things.

One question out of personal interest: Did you find a good setup with Claude Code to write actual VHDL production code? Do you have specific skill files or the like in use?

My success for VHDL production code still is limited. I had cases were I got usable code - but very often I got code that was far from usable and in the end costed me more time to debug than write myself. So I'd like to know if on that front there is something I can learn from you.

For test-benches, documentation, boilerplate code and things like making code linter compliant I was successful already. It's mainly the actual business logic I still had issues finding a really usable setup.

Remark: This comment does **NOT** in any way mean I dislike you are using it. I don't care if good code is manually written or by AI - and you properly disclosed it. It really is about myself being happy about AI inputs from other HDL designers.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
```diff
@@ -0,0 +1,282 @@
...
+        WrEna  <= '1';
+        WrFlip <= flipBits;
+        wait until rising_edge(Clk);
+        WrEna  <= '0';
+        WrFlip <= (WrFlip'range => '0');
+        Addr   <= toUslv(0, Addr'length);
+        WrData <= toUslv(0, WrData'length);
+    end procedure;
+
+    procedure checkEcc (
```
obruendl, line 103: This function does not allow checking back to back reads.

I know that my own RAM TBs are old-fashioned and not very good style (I portes ome of them from old non-vunit code). But this does not mean we cannot do better here.

Did you consider using AXI-S VCs (a master for executing the read/write commands and a slave for checking the output)? This would allow executing back-to-back transactions and playing with random timing.

It is not a 100% absolute must - but I think it would lead to better TB code quality (and potentially also shorter code).

I already commented in the entity that it might be good to add an _RdValid_ signal. If you do so, the usage of AXI4-S VCs would become trivial because the valid signal already is there.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
```diff
@@ -0,0 +1,282 @@
...
+    -----------------------------------------------------------------------------------------------
+    -- Constants
+    -----------------------------------------------------------------------------------------------
+    constant ClkPeriod_c     : time     := 10 ns;
+    constant CodewordWidth_c : positive := eccCodewordWidth(Width_g);
+
+    -----------------------------------------------------------------------------------------------
+    -- Bit-flip pattern helpers
+    -----------------------------------------------------------------------------------------------
+    function singleBit (idx : natural) return std_logic_vector is
```
obruendl: Some functions like these you might want to move to packages.

I think a function `setBits(indexes : IntegerArray_t, width: positive) return std_logic_vector` would be a nice addition to _olo_base_pkg_logic_.

This way you can use it in all your Test-Benches - and it could be useful in production code or for other uses as well.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
```diff
@@ -0,0 +1,282 @@
...
+    function doubleBit (idxA : natural; idxB : natural) return std_logic_vector is
+        variable Result_v : std_logic_vector(CodewordWidth_c - 1 downto 0) := (others => '0');
+    begin
+        Result_v(idxA) := '1';
+        Result_v(idxB) := '1';
+        return Result_v;
+    end function;
+
+    -----------------------------------------------------------------------------------------------
+    -- TB Definitions
```
obruendl, line 45: I think these functions will be sued for all _olo_ft_ RAM implementations. You could move them to a shared package.

obruendl (commented):
I forgot to review the TB in the review before.

src/ft/vhdl/olo_ft_pkg_ecc.vhd
rustyqt: Agreed, using isPower2() from olo_base_pkg_math instead

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: Accepted and applied. The generic is RamRdLatency_g everywhere now — entity ports, testbench, doc, tools/inference_test/yaml/ft.yml and sim/test_configs/olo_ft.py. The doc explicitly states total read latency = RamRdLatency_g + EccPipeline_g.

src/ft/vhdl/olo_ft_pkg_ecc.vhd
rustyqt: Done. test/ft/olo_ft_pkg_ecc/olo_ft_pkg_ecc_tb.vhd covers:

direct value checks for eccParityBits / eccCodewordWidth at widths 8 / 16 / 32 / 64 / 128
encode -> decode round-trip
SecAllBits-w{8,16,32,64,128} — single-bit flip at every codeword position, asserting SEC + correction
DedAllPairs-w8 — exhaustive 78-pair sweep
DedSampledPairs-w64 / -w128 — representative pairs (parity+parity, parity+data, far-apart, adjacent)

doc/ft/olo_ft_ram_sp.md
rustyqt: Done. New doc/ft/olo_ft_principles.md covers: Why Fault Tolerance, SECDED Hamming Code (codeword layout + ECC overhead), Error Injection (popcount table), Error Status Flags, Instantiation from Verilog and VHDL, Read-Data Valid, ECC Pipeline, and the constraints that apply across the area. olo_ft_ram_sp.md and olo_ft_pkg_ecc.md reference it instead of duplicating shared content; the ## ft intro in EntityList.md links to it.

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: Agreed and added. olo_ft_ram_sp gains a RdValid output that pulses '1' on the cycle when RdData / RdEccSec / RdEccDed correspond to a user-issued read. Implementation is a registered shift-register that delays not WrEna by RamRdLatency_g + EccPipeline_g, so the data path and the valid pipe are time-aligned regardless of EccPipeline_g.

doc/ft/olo_ft_ram_sp.md
rustyqt: Done. The WrEccBitFlip row in the Interfaces table now links to the eccCodewordWidth anchor in the package doc, and the default is written as "all 0" instead of (others => '0'). The same change applied to the Error Injection prose.

doc/ft/olo_ft_ram_sp.md
rustyqt: Done. New doc/ft/olo_ft_ram_sp_arch.svg embedded in the Architecture section.

doc/ft/olo_ft_pkg_ecc.md
rustyqt: Done. The doc has a Functions section with one heading per function (eccParityBits, eccCodewordWidth, eccEncode, eccSyndromeAndParity, eccCorrectData, eccSecError, eccDedError); each entry has a VHDL signature block and a short description. Also added a per-function **Usage:** tag separating elaboration-time helpers from combinational RTL, none of them are testbench-only.

src/ft/vhdl/olo_ft_pkg_ecc.vhd
rustyqt: Two new entities:

olo_ft_ecc_encode (generics Width_g, Pipeline_g) — combinational SECDED encode with the codeword-wide In_BitFlip injection input you suggested. Pipeline_g=0 is combinational; >0 adds register stages.
olo_ft_ecc_decode (generics Width_g, Pipeline_g) — combinational SECDED decode emitting Out_Data / Out_EccSec / Out_EccDed, all time-aligned. The pipeline registers that previously lived inside olo_ft_ram_sp are now inside this entity.

olo_ft_ram_sp is refactored to instantiate them — the inlined codec body and the EccPipeline_g generate block are gone. Each new entity has its own VUnit testbench (swept across Width_g, Pipeline_g) and its own Markdown doc.

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: Done. The two new entities provide pipelining, as described in comment above.

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: Thanks for raising this early. Answers:

1. Generic vs separate entity. Leaning toward keeping olo_ft_ram_sp_scrub as a separate entity (the prototype already has one). A Scrubbing_g boolean on olo_ft_ram_sp would drag the scrubber's Rst / Scrub_Stop / Scrub_Stopped / Scrub_Active / Scrub_Addr / Scrub_EccSec / Scrub_EccDed / Scrub_PassDone ports into the no-scrub configuration.

2. Scheduling modes. Prototype currently implements only the "scrubber priority via explicit handshake" pattern (Scrub_Stop / Scrub_Stopped), with that the scrubbing rate is user controlled. Other autonomous modes are valuable but I'd add them incrementally via a ScrubMode_g generic once the baseline is in. TDP-port-split for user and scrubber was considered, but came with a tone of additional problems on avoiding conflicting writes between the write ports with async clock domains.

3. Scrubber as a private entity. Agree. Plan is to extract an internal olo_ft_ram_scrubber (private, not in EntityList) that the public _scrub wrappers instantiate, removing the duplicated FSM logic between sp_scrub and sdp_scrub.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
rustyqt: I agree that we can do better than that. For advanced verification techniques I would like to add UVVM support for open-logic modules. For this I have created a new feature request: https://github.com/open-logic/open-logic/issues/309 

I suggest to discuss the pro's and con's of adding UVVM VVC vs utilizing VUnit VCs over there.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
rustyqt: Moved to olo_base_pkg_logic

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
rustyqt: > Done — answering together because they collapse into one change.
>
> Three `setBits` overloads added to `olo_base_pkg_logic`:
>
> ```vhdl
> function setBits (indexes : in IntegerArray_t; width : in positive) return std_logic_vector;
> function setBits (idx     : in natural;        width : in positive) return std_logic_vector;
> function setBits (idxA    : in natural; idxB  : in natural; width : in positive) return std_logic_vector;
> ```
>
> `olo_base_pkg_logic_tb` gains a `setBits` test case.

doc/ft/olo_ft_ram_sp.md
rustyqt: Your argumentation convinced me. This will become very handy when it comes to hardware level validation. Implemented as suggested.

WrEccBitFlip is replaced by two ports:                                                                    

- ErrInj_BitFlip — codeword-wide flip pattern
- ErrInj_Valid — strobe that latches the current ErrInj_BitFlip value into an internal pending register

The latched pattern is applied to the next write and cleared after that write. If ErrInj_Valid and WrEna are both '1' in the same cycle the pattern is applied immediately.

rustyqt:
Regarding AI usage I have developed a setup based on a set of skills. These skills refer to development guidelines as well as applicable industry standards. In principle the skill for fpga module development defines a six phase approach which is mandatory for every new module development:

1. Write Requirements specification (flow-down from user-provided requirements file + module design requirements)
2. Architecture Definition Report (write architectural approach to a document for user review)
3. Verification Plan (write verification approach for user review - the skill already specifies the verification frameworks to be used, i.e. VUnit + UVVM)
4. Implement RTL
5. Implement VUnit + UVVM based Testbench
6. Verify (Execute VUnit tests until all bugs are fixed)
7. Additional Documents: Verification Report + Design Report (optional)

Each step is approved by user review unless otherwised requested. Extremely adviceable for complex modules to avoid the AI agent going in a wrong direction early on.

Furthermore, for advanced debugging I have an skill that allows the AI to querry waveform data on its own to find the root cause of an identified bug. I'm still optimizing this skill, but the idea is to provide an AI native for looking at waveforms.

src/ft/vhdl/olo_ft_pkg_ecc.vhd
obruendl: OK

src/ft/vhdl/olo_ft_ram_sp.vhd
obruendl: OK so far.

On topic 2: Could we add the "scrubber automatically uses cycles where the user does not interact with the RAM" option?

IMO this would add a whole lot of value and implementation wise it should not be very complex. 

I am fine with not implementing all options yet - but the value of a scrubber that does NOT interfer with the user is big enough to push it forward IMO.

doc/ft/olo_ft_ecc_decode.md
```diff
@@ -0,0 +1,38 @@
...
+_Out_EccDed_ but cannot be corrected. _Out_Data_, _Out_EccSec_ and _Out_EccDed_ are time-aligned.
+
+For background on the SECDED scheme and codeword layout, see
+[Open Logic Fault-Tolerance Principles](./olo_ft_principles.md).
+
+## Generics
+
+| Name       | Type     | Default | Description                                                  |
+| :--------- | :------- | ------- | :----------------------------------------------------------- |
+| Width_g    | positive | -       | Number of data bits per codeword                             |
```
obruendl: Make clear that this is the "Uncoded Data Width" (current wording could be interpreted as "number of bits in codeword")

doc/ft/olo_ft_ecc_encode.md
```diff
@@ -0,0 +1,44 @@
...
+  and friends to avoid duplicating the encoder body).
+- Forward error correction (FEC) over a transmission link.
+- AXI4-Stream traffic that is buffered in unprotected external memory (encode at the producer, decode at the
+  consumer with [olo_ft_ecc_decode](./olo_ft_ecc_decode.md)).
+
+## Generics
+
+| Name       | Type     | Default | Description                                                  |
+| :--------- | :------- | ------- | :----------------------------------------------------------- |
+| Width_g    | positive | -       | Number of data bits per codeword                             |
```
obruendl: Same here: Clarify the wording to say this is "Uncoded data bits" and not "number of bits in codeword".

src/ft/vhdl/olo_ft_pkg_ecc.vhd
obruendl: OK

doc/ft/olo_ft_ecc_decode.md
```diff
@@ -0,0 +1,38 @@
+<img src="../Logo.png" alt="Logo" width="400">
+
+# olo_ft_ecc_decode
+
+[Back to **Entity List**](../EntityList.md)
+
+## Status Information
```
obruendl, line 7: The coverage badges are missing.

Simply add them according to the pattern for the other entities in open-logic. They will be generated automatically when the pipeline runs coverage tests on _olo_ft_ for the first time.

doc/ft/olo_ft_ecc_encode.md
```diff
@@ -0,0 +1,44 @@
+<img src="../Logo.png" alt="Logo" width="400">
+
+# olo_ft_ecc_encode
+
+[Back to **Entity List**](../EntityList.md)
+
+## Status Information
```
obruendl, line 7: The coverage badges are missing.

Simply add them according to the pattern for the other entities in open-logic. They will be generated automatically when the pipeline runs coverage tests on _olo_ft_ for the first time.

doc/ft/olo_ft_pkg_ecc.md
```diff
@@ -0,0 +1,126 @@
+<img src="../Logo.png" alt="Logo" width="400">
+
+# olo_ft_pkg_ecc
+
+[Back to **Entity List**](../EntityList.md)
+
+## Status Information
```
obruendl, line 7: The coverage badges are missing.

Simply add them according to the pattern for the other entities in open-logic. They will be generated automatically when the pipeline runs coverage tests on _olo_ft_ for the first time.

doc/ft/olo_ft_ram_sp.md
```diff
@@ -0,0 +1,78 @@
+<img src="../Logo.png" alt="Logo" width="400">
+
+# olo_ft_ram_sp
+
+[Back to **Entity List**](../EntityList.md)
+
+## Status Information
```
obruendl, line 7: The coverage badges are missing.

Simply add them according to the pattern for the other entities in open-logic. They will be generated automatically when the pipeline runs coverage tests on _olo_ft_ for the first time.

doc/ft/olo_ft_ecc_decode.md
```diff
@@ -0,0 +1,38 @@
...
+[Open Logic Fault-Tolerance Principles](./olo_ft_principles.md).
+
+## Generics
+
+| Name       | Type     | Default | Description                                                  |
+| :--------- | :------- | ------- | :----------------------------------------------------------- |
+| Width_g    | positive | -       | Number of data bits per codeword                             |
+| Pipeline_g | natural  | 0       | Number of register stages after the decode combinational logic. 0 = combinational output. |
+
+## Interfaces
```
obruendl, line 54: For consistency with other entities, please split into 3 tables (control signals, input, output).

src/ft/vhdl/olo_ft_ecc_decode.vhd
```diff
@@ -0,0 +1,102 @@
...
+-- Entity
+---------------------------------------------------------------------------------------------------
+entity olo_ft_ecc_decode is
+    generic (
+        Width_g    : positive;
+        Pipeline_g : natural := 0
+    );
+    port (
+        Clk         : in    std_logic                                                  := '0';
+        In_Codeword : in    std_logic_vector(eccCodewordWidth(Width_g) - 1 downto 0);
```
obruendl: Can we add _In_Valid_ and _Out_Valid_? Or course this requires also to have _Rst_ to reset the valid registers.

For the usage in cases like FEC on data-links having at least AXI4-Stream forward handshaking would be great. Not having backpressure (_Ready_) is acceptable IMO.

src/ft/vhdl/olo_ft_ecc_encode.vhd
```diff
@@ -0,0 +1,83 @@
...
+-- Entity
+---------------------------------------------------------------------------------------------------
+entity olo_ft_ecc_encode is
+    generic (
+        Width_g    : positive;
+        Pipeline_g : natural := 0
+    );
+    port (
+        Clk          : in    std_logic                                                  := '0';
+        In_Data      : in    std_logic_vector(Width_g - 1 downto 0);
```
obruendl: Can we add _In_Valid_ and _Out_Valid_? Or course this requires also to have _Rst_ to reset the valid registers.

For the usage in cases like FEC on data-links having at least AXI4-Stream forward handshaking would be great. Not having backpressure (_Ready_) is acceptable IMO.

doc/ft/olo_ft_ecc_encode.md
```diff
@@ -0,0 +1,44 @@
...
+  consumer with [olo_ft_ecc_decode](./olo_ft_ecc_decode.md)).
+
+## Generics
+
+| Name       | Type     | Default | Description                                                  |
+| :--------- | :------- | ------- | :----------------------------------------------------------- |
+| Width_g    | positive | -       | Number of data bits per codeword                             |
+| Pipeline_g | natural  | 0       | Number of register stages after the encode/inject combinational logic. 0 = combinational output. |
+
+## Interfaces
```
obruendl, line 53: For consistency with other entities, please split into 3 tables (control signals, input, output).

doc/ft/olo_ft_ram_sp_arch.svg
obruendl, line 1: Please move the figure into a separate folder. I want exactly one file per entity in the "area" folder and figures in subfolders - otherwise the "area" folder gets flooded when a few entities with some figures each are present.

obruendl, line 1: Can you make it a *.drawio.png (with the drawio data for editing it included)? Please ensure white background for readability on browsers with dark-theme - not sure if this was present in the SVG.

doc/ft/olo_ft_ram_sp.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Interfaces
+
+| Name         | In/Out | Length                | Default | Description                                                  |
+| :----------- | :----- | :-------------------- | ------- | :----------------------------------------------------------- |
+| Clk          | in     | 1                     | -       | Clock                                                        |
+| Addr         | in     | _ceil(log2(Depth_g))_ | -       | Address                                                      |
+| WrEna        | in     | 1                     | '1'     | Write enable                                                 |
+| WrData       | in     | _Width_g_             | -       | Write data                                                   |
+| ErrInj_BitFlip | in   | _[eccCodewordWidth](./olo_ft_pkg_ecc.md#ecccodewordwidth)(Width_g)_ | all 0 | ECC error-injection pattern (codeword-wide). Each '1' bit XORs (flips) the corresponding bit of the stored codeword. Popcount 1 = SEC-correctable, popcount 2 = DED-detectable. Latched into an internal register when _ErrInj_Valid_ = '1' and applied to the next write. |
```
obruendl: Can we move the error injection into a separate section within interface?

Core idea: Like that we have the main functionality the user for sure wants to sue separated from error injection which is optional.

doc/EntityList.md
```diff
@@ -31,6 +31,9 @@ Note that components are split into categories.
     - [CORDIC](#cordic)
     - [CIC Filters](#cic-filters)
     - [Miscellaneous](#miscellaneous-1)
+  - [ft](#ft)
+    - [Packages (olo\_ft\_pkg\_\<...\>)](#packages-olo_ft_pkg_)
```
obruendl, line 36: Link to CODEC section is missing

doc/EntityList.md
```diff
@@ -256,3 +259,30 @@ be used. For deciding which option to use, the following considerations shall be
...
+
+The cross-cutting concepts (codeword layout, ECC overhead, error injection semantics, status flags, ECC pipeline,
+common constraints) are described once in
+[Open Logic Fault-Tolerance Principles](./ft/olo_ft_principles.md) and referenced from the per-entity docs.
+
+### Packages (olo_ft_pkg_\<...\>)
+
+| Package                                       | Description                                                  |
+| --------------------------------------------- | ------------------------------------------------------------ |
+| [olo_ft_pkg_ecc](./ft/olo_ft_pkg_ecc.md)      | SECDED Hamming code functions for ECC-protected memories. **For internal use within Open Logic only** |
```
obruendl: Is there a specific reason why users are not allowed to use the package?

I f not, I'd remove the comment. The package seems fine to be used by users if they want to have functions IMO.

doc/EntityList.md
```diff
@@ -256,3 +259,30 @@ be used. For deciding which option to use, the following considerations shall be
...
+
+### ECC Codec (olo_ft_ecc_\<...\>)
+
+| Entity                                          | Description                                                  |
+| ----------------------------------------------- | ------------------------------------------------------------ |
+| [olo_ft_ecc_encode](./ft/olo_ft_ecc_encode.md)  | SECDED encoder with optional pipeline and bit-flip injection (Verilog-callable) |
+| [olo_ft_ecc_decode](./ft/olo_ft_ecc_decode.md)  | SECDED decoder with optional pipeline (Verilog-callable)     |
```
obruendl: Remove the "Verilog-callable". This is the standard for open-logic. 

Reason: If we start adding it here, we have to add it to all entities because users otherwise think entities that do not have the comment might not be verilog compatible.

doc/EntityList.md
```diff
@@ -256,3 +259,30 @@ be used. For deciding which option to use, the following considerations shall be
...
+
+### RAM Implementations (olo_ft_ram_\<...\>)
+
+| Entity                                   | Description                                                  |
+| ---------------------------------------- | ------------------------------------------------------------ |
+| [olo_ft_ram_sp](./ft/olo_ft_ram_sp.md)   | ECC-protected single port RAM (SECDED Hamming)               |
```
obruendl: I'd just say "ECC-rpotected single port RAM". 

If you add SECDED here you have to add it for all entities. And the introductory section says that we do SECDED ECC. So we do not have to repeat ourselves.

sim/test_configs/olo_ft.py
```diff
@@ -0,0 +1,39 @@
...
+        for Pipeline in [0, 1, 2]:
+            named_config(tb, {'Pipeline_g': Pipeline})
+
+    ### olo_ft_ram_sp ###
+    tb = olo_tb.test_bench('olo_ft_ram_sp_tb')
+    for RamBehav in ['RBW', 'WBR']:
+        named_config(tb, {'RamBehavior_g': RamBehav})
+    for ReadLatency in [1, 2]:
+        named_config(tb, {'RamRdLatency_g': ReadLatency})
+    for Width in [8, 16, 32, 64]:
```
obruendl: I think 2 powers of two are sufficient. But Adding a non-power-of-two would be good - this really is something different.

sim/test_configs/olo_ft.py
```diff
@@ -0,0 +1,39 @@
...
+def add_configs(olo_tb):
+    """
+    Add all fault-tolerant testbench configurations to the VUnit Library
+    :param olo_tb: Testbench library
+    """
+
+    ### olo_ft_ecc_encode / olo_ft_ecc_decode ###
+    for tb_name in ['olo_ft_ecc_encode_tb', 'olo_ft_ecc_decode_tb']:
+        tb = olo_tb.test_bench(tb_name)
+        for Width in [8, 16, 32, 64]:
```
obruendl: I suggest only having two powers of two but one other case (e.g. 13 bits) - to see if this works correctly as well.

sim/run.py
```diff
@@ -130,7 +130,10 @@ class Simulator(Enum):
 
 # Disable IEEE numeric_std warnings to avoid slow simulation speed due to X-propagation
 olo_tb.set_sim_option('ghdl.sim_flags', ['--ieee-asserts=disable'])
-olo_tb.set_sim_option('nvc.global_flags', ['--ieee-warnings=off'])
+try:
+    olo_tb.set_sim_option('nvc.global_flags', ['--ieee-warnings=off'])
+except ValueError:
+    pass  # nvc.global_flags not supported in older VUnit versions
```
obruendl: Can we print the error clearly as "IGNORED ERROR:..." or warning or the likes and say that it is ignored because expected for older VUnit versions - and suggest to the user to install a later one?

I don't feel good simply suppressing all _ValueErrors_. Could well be that there are other reasons to raise a _ValueError_ and supressing them would make debugging hard.

src/base/vhdl/olo_base_pkg_logic.vhd
```diff
@@ -74,6 +75,12 @@ package olo_base_pkg_logic is
 
     function getTrailingSetBitIndex (vec : in std_logic_vector) return integer;
 
+    -- Build a std_logic_vector of given width with bits set at the given indexes,
+    -- all other bits cleared. Useful for constructing bit-flip patterns or masks.
+    function setBits (indexes : in IntegerArray_t; width : in positive) return std_logic_vector;
+    function setBits (idx     : in natural;        width : in positive) return std_logic_vector;
+    function setBits (idxA    : in natural; idxB  : in natural; width : in positive) return std_logic_vector;
```
obruendl: Why do we need the second and third line? I think the first line can be used to represent those cases as well:

```
setBits((5), 10); -- replaces second line
setBits((3,5), 10); -- replaced third line
```

Or did I miss anything?

obruendl: I think the new functions must be added to the documentation of _olo_base_pkg_logic_ as well.

src/ft/vhdl/olo_ft_ecc_decode.vhd
```diff
@@ -0,0 +1,102 @@
...
+
+    -- Pipeline: register stages after decode
+    g_pipe : if Pipeline_g > 0 generate
+        type Data_t is array (natural range <>) of std_logic_vector(Width_g - 1 downto 0);
+        signal DataPipe : Data_t(1 to Pipeline_g);
+        signal SecPipe  : std_logic_vector(1 to Pipeline_g);
+        signal DedPipe  : std_logic_vector(1 to Pipeline_g);
+        attribute shreg_extract of DataPipe : signal is ShregExtract_SuppressExtraction_c;
+        attribute shreg_extract of SecPipe  : signal is ShregExtract_SuppressExtraction_c;
+        attribute shreg_extract of DedPipe  : signal is ShregExtract_SuppressExtraction_c;
```
obruendl: Also add 

```
        attribute syn_srlstyle of VldReg  : signal is SynSrlstyle_FlipFlops_c;
        attribute syn_srlstyle of DataReg : signal is SynSrlstyle_FlipFlops_c;
``

File: src/ft/vhdl/olo_ft_ecc_decode.vhd [hunk 35]
```diff
@@ -0,0 +1,102 @@
...
+
+begin
+
+    -- Combinational SECDED decode
+    SynPar   <= eccSyndromeAndParity(In_Codeword, Width_g);
+    DataCorr <= eccCorrectData(In_Codeword, SynPar, Width_g);
+    EccSecI  <= eccSecError(SynPar);
+    EccDedI  <= eccDedError(SynPar);
+
+    -- No pipeline: combinational output
```
obruendl: IMO it would make sense to describe the logic to describe the explicitly distributed over the two register stages instead of relying on register retiming. Tools do often fail on retiming, so we better should describe the logic already properly registered for _Pipeline_g>1_:

```
...
if rising_edge(Clk) then
   -- Stage 0
   SynPar_0 <= eccSyndromeAndParity(In_Codeword, Width_g);
   In_Codeword_0 <= In_Codeword;
   -- Stage 1
   DataCorr_1 <= EccCorrectData(In_Codeword_0, SynPar_0, Width_g);
   EccSecI_1 <= eccSecError(SynPar_0);
   EccDedI_1 <= eccDedError(SynPar_0);
...
```

src/ft/vhdl/olo_ft_ecc_encode.vhd
```diff
@@ -0,0 +1,83 @@
...
+        In_BitFlip   : in    std_logic_vector(eccCodewordWidth(Width_g) - 1 downto 0)   := (others => '0');
```
obruendl: I'd suggest to ave error injection asynchronous to the normal data (with its separate valid signal) land. What we discussed for the RAM (that the error injection might be requested separately fro mthe data-flow timing) is valid for the encode entity as well.

src/ft/vhdl/olo_ft_ram_sp.vhd
```diff
@@ -0,0 +1,152 @@
...
+
+    p_pending : process (Clk) is
+    begin
+        if rising_edge(Clk) then
+            if WrEna = '1' then
+                -- Pattern (if any) was applied to this write; clear so it isn't re-applied.
+                ErrInj_Pending <= (others => '0');
+            elsif ErrInj_Valid = '1' then
+                -- Preload for a future write.
+                ErrInj_Pending <= ErrInj_BitFlip;
```
obruendl: `ErrInj_Pending` requires a reset. Otherwise an error injection requested before the reset may be executed after the reset, which is violating the idea of fully resetting the state.

I suggested to move this into encode/decode anyways. But I make the comment still - because the reset must be added when the logic is moved there.

src/ft/vhdl/olo_ft_ecc_decode.vhd
```diff
@@ -0,0 +1,102 @@
...
+
+---------------------------------------------------------------------------------------------------
+-- Entity
+---------------------------------------------------------------------------------------------------
+entity olo_ft_ecc_decode is
+    generic (
+        Width_g    : positive;
+        Pipeline_g : natural := 0
+    );
+    port (
```
obruendl, line 44: Why does the _decode_ entity not have error injectinon when the _encode_ entity does?

I understand that for the RAM you only need it on the _encode_ side. But if users use _encode_ and _decode_ separately (e.g. to ECC protect a data stream), they want error injection on both sides I'd expect.

src/ft/vhdl/olo_ft_ram_sp.vhd
```diff
@@ -0,0 +1,152 @@
...
+        port map (
+            Clk         => Clk,
+            In_Codeword => Rd_Codeword,
+            Out_Data    => RdData,
+            Out_EccSec  => RdEccSec,
+            Out_EccDed  => RdEccDed
+        );
+
+    -- Read-valid pipeline: a read happens whenever WrEna='0'. Delay matches data path.
+    p_rd_valid : process (Clk) is
```
obruendl: If we add the _Valid_ to the _encode_ and _decode_ entities (as suggested there) we do not need the process anymore because _olo_ft_ecc_decode_ natively comes with the _Valid_ signal.

test/ft/olo_ft_ecc_decode/olo_ft_ecc_decode_tb.vhd
```diff
@@ -0,0 +1,121 @@
...
+            Out_Data    => Out_Data,
+            Out_EccSec  => Out_EccSec,
+            Out_EccDed  => Out_EccDed
+        );
+
+    test_runner_watchdog(runner, 1 ms);
+
+    p_control : process is
+
+        procedure applyAndCheck (
```
obruendl: The problem with the _applyAndCheck_ function is that it always injects only one sample and then waits until the result is there. It never checks if the  logic works fine if data is processed back-to-back.

I didn't see anything suspicious in your code - but tests also serve as protection against regressions. Currently we would not detect if pipeline stages are mixed in the implementation.

As soon as we have the _Valid_ signal as I requested, you can use the _axi_stream_master_t_ and _axi_strea_slave_t_ verification components provided by VUnit. This allows you to simulate random timing, back to back data, etc.

I'd suggest to go this way. For an example have a look at the _olo_base_wconv_n2m_ testbench.

You can simply add a _push_axi_stream()_ and a _check_axi_stream()_ function into _applyAndCheck()_ so you can achieve the new setup without having even to change the test-cases. _Sed_ and _Ded_ I'd putt into tuser.

test/ft/olo_ft_ecc_encode/olo_ft_ecc_encode_tb.vhd
```diff
@@ -0,0 +1,118 @@
...
+            In_BitFlip   => In_BitFlip,
+            Out_Codeword => Out_Codeword
+        );
+
+    test_runner_watchdog(runner, 1 ms);
+
+    p_control : process is
+        variable Expected_v : std_logic_vector(CodewordWidth_c - 1 downto 0);
+
+        procedure applyAndCheck (
```
obruendl: See comment in decode entity. I'd suggest using the VUnit AXI-Stream VCs, so you can emulate back-to-back data processing and random timing.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
```diff
@@ -0,0 +1,320 @@
...
+        WrData <= toUslv(data, WrData'length);
+        WrEna  <= '1';
+        wait until rising_edge(Clk);
+        WrEna  <= '0';
+        Addr   <= toUslv(0, Addr'length);
+        WrData <= toUslv(0, WrData'length);
+    end procedure;
+
+    -- Write with simultaneous error injection (legacy semantics: flip applied immediately).
+    procedure writeWithFlip (
```
obruendl: _write_ and _writeWithFlip_ are almost identical. IMO you could put both into one _write_ function and use a default value "0" for _flipBits_. If the default value is not changed by the user, do not asser _InjValid_, otherwise do assert it.

Like that you can get rid of one copy of the funciton.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
```diff
@@ -0,0 +1,320 @@
...
+            wait until rising_edge(Clk);
+        end loop;
+
+        check_equal(RdValid, '1',                          message & " RdValid");
+        check_equal(RdData,  toUslv(data, RdData'length),  message & " data");
+        check_equal(EccSec,  expEccSec,                    message & " EccSec");
+        check_equal(EccDed,  expEccDed,                    message & " EccDed");
+    end procedure;
+
+    procedure checkDedOnly (
```
obruendl: Similar to above, the check function is twice here and almost identical. By adding a _IgnoreData_ or _CheckData_ argument to enable/disable the checking of data, you could avoid duplicating it.

tools/inference_test/yaml/ft.yml
```diff
@@ -0,0 +1,118 @@
...
+exclude_entities:
+  - "olo_private_*"
+  - "olo_base_*"
+
+entities:
+
+  # ============================================================
+  # olo_ft_ecc_encode
+  # ============================================================
+  - entity_name: "olo_ft_ecc_encode"
```
obruendl, line 15: Please keep only one config per enttiy in ft.yml. 

The test synthesis takes significant time to execute on the AWS runners. I'd like to avoid that the pipeline take days - which it would if we add several configurations for every component.

If you really want to preserve a YAML with many configurations, we should create a separate file. Do you want to preserve it or is one configuration sufficient for you?

compile_order.txt
obruendl, line 1: You have to regenerate this one, otherwise the yosys test compilation (colgneChip) for the encode/decode entities will fail (they are not yet in the compile order).

obruendl:
> Regarding AI usage I have developed a setup based on a set of skills. These skills refer to development guidelines as well as applicable industry standards. In principle the skill for fpga module development defines a six phase approach which is mandatory for every new module development:
> 
> 1. Write Requirements specification (flow-down from user-provided requirements file + module design requirements)
> 2. Architecture Definition Report (write architectural approach to a document for user review)
> 3. Verification Plan (write verification approach for user review - the skill already specifies the verification frameworks to be used, i.e. VUnit + UVVM)
> 4. Implement RTL
> 5. Implement VUnit + UVVM based Testbench
> 6. Verify (Execute VUnit tests until all bugs are fixed)
> 7. Additional Documents: Verification Report + Design Report (optional)
> 
> Each step is approved by user review unless otherwised requested. Extremely adviceable for complex modules to avoid the AI agent going in a wrong direction early on.
> 
> Furthermore, for advanced debugging I have an skill that allows the AI to querry waveform data on its own to find the root cause of an identified bug. I'm still optimizing this skill, but the idea is to provide an AI native for looking at waveforms.

Is your skill-set proprietary and protected or would you be willing to share? Or do you already publish them somewhere anyways?

Especially the waveform skill I'd be very intersted in.

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: With the single port RAM interface automatic scrubbing is not feasible as it is not possible to tell when the RAM is used and when not. (Wr_Ena = '1' --> Write Access, Wr_EnA = '0' --> Read Access). This makes a handshake mechanism mandatory for the single port RAM.

In this case I would even argue to omit the scrubbing option for the the single port RAM entirely and implement the _Srub_g_ option for the sdp RAM which enables automatic scrubbing. This would move the topic to the next PR.

doc/ft/olo_ft_ecc_decode.md
rustyqt: Agreed. Changed to: "Number of user data bits"

doc/ft/olo_ft_ecc_encode.md
rustyqt: Agreed. Changed to: "Number of user data bits"

compile_order.txt
rustyqt: OK, done — regenerated via `python sim/run.py --compile_list`. Encode/decode entries now appear between `olo_ft_pkg_ecc.vhd` and `olo_ft_ram_sp.vhd`. Also normalised the path separators to forward slashes in `sim/run.py` so the file stays identical between Windows and Linux runs.

doc/ft/olo_ft_ecc_decode.md
rustyqt: OK, done. Three-badge block (coverage / branches / issues) added matching the pattern from `olo_base_ram_sp.md`.

doc/ft/olo_ft_ecc_encode.md
rustyqt: OK, done.

doc/ft/olo_ft_pkg_ecc.md
rustyqt: OK, done.

doc/ft/olo_ft_ram_sp.md
rustyqt: OK, done.

doc/ft/olo_ft_ecc_decode.md
rustyqt: OK, done. Split into Clock-and-Reset / Input / Output, plus a separate Error-Injection table for the optional injection ports.

doc/ft/olo_ft_ecc_encode.md
rustyqt: OK, done.

src/ft/vhdl/olo_ft_ecc_decode.vhd
rustyqt: OK, done. Decoder now exposes In_Valid / In_Ready / Out_Valid / Out_Ready (AXI4-Stream forward handshake) plus Rst. Internally backed by `olo_base_pl_stage` (with a `UseReady_g` generic so the RAM/FIFO wrappers can opt out of the shadow-register area).

src/ft/vhdl/olo_ft_ecc_encode.vhd
rustyqt: OK, done — same shape as the decoder.

doc/ft/olo_ft_ram_sp_arch.svg
rustyqt: OK, done. Diagram now lives at `doc/ft/ram/olo_ft_ram_sp_arch.drawio.png`.

rustyqt: OK, done. The PNG embeds the full diagram XML in a `mxfile` tEXt chunk, so drawio can re-open and edit it directly. White background guaranteed.

doc/ft/olo_ft_ram_sp.md
rustyqt: OK, done. The two `ErrInj_*` ports moved to their own subsection under Interfaces.

doc/EntityList.md
rustyqt: OK, done. ToC entry for the new ECC Codec subsection added.

doc/EntityList.md
rustyqt: OK, done — no real reason to gate the package, removed the comment.

doc/EntityList.md
rustyqt: OK, done.

doc/EntityList.md
rustyqt: OK, done — now just "ECC-protected single port RAM".

sim/test_configs/olo_ft.py
rustyqt: OK, done. Width sweep is now `[8, 13, 32]` (two powers of two + 13 as the non-power-of-two).

sim/test_configs/olo_ft.py
rustyqt: OK, done — same `[8, 13, 32]` sweep applied to the encode/decode TBs as well.

sim/run.py
rustyqt: OK, done. The handler now prints an explicit `IGNORED: nvc.global_flags not recognised by this VUnit version (...). This is expected on older VUnit installations; upgrade VUnit to enable nvc IEEE-warning suppression.` and only catches `ValueError`, so other exceptions still surface.

src/base/vhdl/olo_base_pkg_logic.vhd
rustyqt: Partial pushback: in VHDL, `(5)` is a parenthesized integer expression, not a single-element aggregate (you'd need `(0 => 5)`). Removing the `setBits(idx : natural; ...)` overload would force every single-bit caller to write `setBits((0 => 5), 10)` which is ugly. The third overload `setBits(idxA, idxB, width)` is genuinely redundant though — your `setBits((3,5), 10)` example works via the array form, so I dropped that one. Result: two overloads — the `IntegerArray_t` form for ≥2 indexes and the single-natural form for the ergonomic 1-index case. Call sites updated.

rustyqt: OK, done. New `### setBits()` block added to `doc/base/olo_base_pkg_logic.md` with the surviving signatures and a small example.

src/ft/vhdl/olo_ft_ecc_decode.vhd
rustyqt: Implicitly satisfied: the codec pipelines now use `olo_base_pl_stage` instances which already set `syn_srlstyle = SynSrlstyle_FlipFlops_c` on their internal valid + data registers, so no manual attribute work is needed in the codec entities themselves.

src/ft/vhdl/olo_ft_ecc_decode.vhd
rustyqt: OK, done. Decoder now has three explicit branches: `Pipeline_g=0` combinational; `Pipeline_g=1` combinational decode + register near the output; `Pipeline_g=2` combinational syndrome → register `{InjectedCw, SynPar}` → combinational correction → register `{Out_Data, Out_EccSec, Out_EccDed}`. Documented in `doc/ft/olo_ft_ecc_decode.md` under "Pipeline structure". `Pipeline_g` now range limited from 0 to 2.

src/ft/vhdl/olo_ft_ecc_encode.vhd
rustyqt: OK, done. Encode now exposes the same paired `ErrInj_BitFlip` + `ErrInj_Valid` interface as the decoder, with the latched-strobe semantics — independent of the data handshake, cleared on `Rst` or on a completed handshake.

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: OK, done. The latch moved into both codec entities so it now has access to `Rst`. It is cleared on `Rst='1'` and on a completed input handshake (`In_Valid AND In_Ready`), so an injection request issued before reset cannot leak into post-reset operation. The RAM wrapper just forwards the strobes.

src/ft/vhdl/olo_ft_ecc_decode.vhd
rustyqt: OK, done. Decoder now exposes the same `ErrInj_BitFlip` + `ErrInj_Valid` interface as the encoder. The pattern is XORed into `In_Codeword` *before* the syndrome calculation, so it simulates a corruption that occurred between the encoder and the decoder (matches the FEC use case you described).

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: OK, mostly done. The decode entity owns the `EccPipeline_g` portion of the read-valid path natively — `decode.Out_Valid` drives the user-facing `RdValid`. The wrapper still has a small shift register of depth `RamRdLatency_g` to delay `(not WrEna)` before feeding `decode.In_Valid`, because `olo_base_ram_sp` itself doesn't expose a read-valid output. Net: the wrapper's helper process shrunk from `RamRdLatency_g + EccPipeline_g` deep to just `RamRdLatency_g` deep, and the EccPipeline_g portion is gone entirely.

test/ft/olo_ft_ecc_decode/olo_ft_ecc_decode_tb.vhd
rustyqt: OK, done. Decode TB now uses `axi_stream_master_t` / `axi_stream_slave_t` from VUnit (with Sec/Ded carried in the slave's `tuser`). `applyAndCheck` replaced by `pushExpect` that calls `push_axi_stream` + `check_axi_stream`. Back-to-back beats are exercised naturally.

test/ft/olo_ft_ecc_encode/olo_ft_ecc_encode_tb.vhd
rustyqt: OK, done — same approach as the decoder TB.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
rustyqt: OK, done. Single `write` procedure with `flipBits : std_logic_vector := (0 downto 0 => '0')` default. The body checks the pattern at runtime — if any bit is `'1'` it asserts `ErrInj_Valid` alongside the write, otherwise it leaves the strobe at `'0'`. The `(0 downto 0 => '0')` default is used because VHDL doesn't accept an unconstrained default that depends on the actual codeword width.

test/ft/olo_ft_ram_sp/olo_ft_ram_sp_tb.vhd
rustyqt: OK, done. Added a `CheckData : boolean := true` argument to `checkEcc`; passing `CheckData => false` skips the data comparison (used in the DED tests where the decoded data is unreliable by definition).

tools/inference_test/yaml/ft.yml
rustyqt: OK, done. Collapsed to one representative config per entity (worst-case width / pipeline depth, with width-reduction on the wide ports to keep synth pin count manageable). I dropped the extra configs entirely rather than moving them to a separate YAML — happy to add a `ft_extended.yml` later if you want broader synthesis coverage outside the CI matrix.

rustyqt:
I have adapted both skills for the use in open-logic. Have a look:

https://github.com/rustyqt/skills

src/ft/vhdl/olo_ft_ram_sp.vhd
```diff
@@ -0,0 +1,156 @@
...
+
+    -- Delay (not WrEna) by RamRdLatency_g to drive decode.In_Valid in lockstep with the
+    -- codeword that just came out of the RAM.
+    p_rd_ena_pipe : process (Clk) is
+    begin
+        if rising_edge(Clk) then
+            if Rst = '1' then
+                RdEnaPipe <= (others => '0');
+            else
+                RdEnaPipe(1) <= not WrEna;
```
obruendl: Can we add a _Rd_Ena_ port (so the user can explicitly state when he wants to read)?

I came accross it due to the scrubbing topic. But I noticed for _RdValid_ it would make sense.

Question: From your perspective, would it make sense to add _Rd_Ena_ and _RdValid_ to the _olo_base_ RAMs as well? It would come at no cost and I just noticed that the _RdValid_ output could help there as well, especially for higher Read Latencies. Here I just want your thought - Implementation I'd take over of course.

src/ft/vhdl/olo_ft_ram_sp.vhd
obruendl: I asked already in another comment: Would it make sense to add _RdValid_ to the _olo_base_ RAMs?

I think your comment indicates this is the case. If I add it in the main develop, you could merge this into your branch and used it. Let me know if you want to go this route, then I quickly add it on develop.

src/ft/vhdl/olo_ft_ram_sp.vhd
obruendl: The topic is for the next PR anyways since I agreed it srubbing RAMs to be separate entities.

I'd say adding a Rd_Ena port to the scrubbing RAMs would make sense - so we know when the user Reads. I noticed that Rd_Ena makes sense for the non-scrubbing RAM as well - so we can reasonably use the RdValid output.

doc/ft/ram/olo_ft_ram_sp_arch.drawio.png
obruendl, line 1: The figure is quite bloated with text. And the arrows should go to the inner blocks (e.g. Error injection arrows should go to the purple block IMO).

I ran into similarly text-bloated figures when I let claude do them. Can you clean it up manually and reduce the text to a level that looks reasonable?

obruendl (commented):
Most comments closed - only a few ones pending.

src/ft/vhdl/olo_ft_ecc_decode.vhd
```diff
@@ -0,0 +1,201 @@
...
+            );
+
+        Out_Data   <= Pl_Out(PayloadWidth_c - 1 downto 2);
+        Out_EccSec <= Pl_Out(1);
+        Out_EccDed <= Pl_Out(0);
+    end generate;
+
+    -- Branch B: Pipeline_g = 2
+    --   Distributed pipeline: stage 1 registers syndrome+parity computation,
+    --   stage 2 registers the corrected data and the SEC/DED flags.
```
obruendl: Can't wevonly have this code and simply use "Stages_g=0" for the unused sedond pipeline stage if the user wants only one register?

Less code - easier maintenance.

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: Honestly, I'm a bit undecided on that one:

- On the one hand, the addition of _RdValid_ would clearly simplify the _olo_ft_ram_sp_ entity
- On the other hand - and this also applies to the current _olo_ft_ram_sp_ interface - the _RdValid_ port is a bit useless without a dedicated _RdEna_ port to make read accesses explicit. Because during idle times, the RAM will be always in read access, this means _RdValid_ will be always high and read data can't be associated with a given input address solely based on the _RdValid_ signal.

In summary, I tend to support the _RdValid_ output port but only if the _olo_base_ram_sp_ and _olo_ft_ram_sp_ entities also get explicit _RdEna_ input ports.

src/ft/vhdl/olo_ft_ram_sp.vhd
rustyqt: @obruendl Now we are on the same page: As I have answered on the previous comment:

- On the one hand, the addition of RdValid would clearly simplify the olo_ft_ram_sp entity
- On the other hand - and this also applies to the current olo_ft_ram_sp interface - the RdValid port is a bit useless without a dedicated RdEna port to make read accesses explicit. Because during idle times, the RAM will be always in read access, this means RdValid will be always high and read data can't be associated with a given input address solely based on the RdValid signal.

In summary, I tend to support the RdValid output port but only if the olo_base_ram_sp and olo_ft_ram_sp entities also get explicit RdEna input ports.

doc/ft/ram/olo_ft_ram_sp_arch.drawio.png
rustyqt: Agreed, I will update the architecture diagram to make it more clean.

src/ft/vhdl/olo_ft_ecc_decode.vhd
rustyqt: Yes, this should work. I will implement it accordingly.

doc/ft/ram/olo_ft_ram_sp_arch.drawio.png
rustyqt: @obruendl I will wait with updating the architecture diagram until we have decided on the interface port additions (_RdValid_ + _RdEna_) for the _olo_base_ram_sp_ .

src/ft/vhdl/olo_ft_ram_sp.vhd
obruendl: I'll add it. I am off this weekend - either I make it tomorrow or it will be next week.

rustyqt: Good, in the meantime i did a "TEMPORARY" commit which can be reverted once I merge the _develop_ branch with the new RAM interfaces to the _feature/fault-tolerant_ branch: https://github.com/open-logic/open-logic/pull/294/changes/74ab2465ac30d747a3a30435ec20ea52fcffe6e8

obruendl: @rustyqt 

I implemented the _RdEna_/_RdValid_ for the _olo_base_ram_sp_.

Please check #315 and let me know if this fits for you (post your comments there).

rustyqt:
@obruendl I'm considering to add optional sideband data for the _olo_ft_ecc_decode_ and _olo_ft_ecc_encode_ entities which won't be encoded/decoded but forwarded with the correct latency. 

This could get useful for...
1. storage information, e.g. a RAM address 
2. routing information, e.g. axi stream tdest signal
3. indicating last word, e.g. axi stream tlast signal (i.e. needed for olo_ft_fifo_packet)

Do you agree?

rustyqt:
@obruendl For your information and transparency on future intends for the ft area, I've created a new branch containing the full backlog of all intended ft features: https://github.com/rustyqt/open-logic/tree/feature/fault-tolerant-all-entities

obruendl:
@rustyqt 

I'd better like not to start adding sideband signals to every entity.

Instead users shall use _olo_base_latency_comp_ to bypass sidebsnd signals alongside a processing chain.

I considered it bu I see the following issues with adding it to every entity:
- violation of DRY principle
- it needs to be tested for every entity (effort)
- for some entities its unclear (would a width converter also width convert sideband data?)

rustyqt:
Okay, agreed. Using _olo_base_latency_comp_ instead.

rustyqt:
Rebased this PR onto current develop to pick up #315 (RdEna/RdValid on the RAMs).

  - Dropped the commit that had the temporary olo_base_ram_sp implementation
  - Regenerated compile_order.txt and applied a couple of minor vsg/markdownlint fixes.

  Verified locally: full GHDL regression green (5301 tests), vsg and markdownlint clean.

src/ft/vhdl/olo_ft_ecc_decode.vhd
rustyqt: Done.

doc/ft/ram/olo_ft_ram_sp_arch.drawio.png
rustyqt: Done. PR #315 was merged to develop containing the RdValid + RdEna on all RAMs. This branch was rebased onto develop once PR #315 was through.

rustyqt:
@obruendl Could you please rebase the _open-logic:feature/fault-tolerant_ branch onto develop as well so the PR diff only contains the ft changes.

obruendl:
@rustyqt 

Thank you for the patience and for fixing everything reported. I think now we are on the right track - I merged the PR to the "feature/fault-tolerant" branch in the main repo where we can integrate the whole _ft_ area.

I think you can now proceed along one or the other path:

a) Implement all the other entities _WITHOUT_ scrubbing. I am pretty sure now they should go through with much less comments becuase we agreed on the general principle. Still I'd prefer having one PR per entity (just to keep discussions scoped).
b) Implement the single-port memory _WITH_ scrubbing.

Both paths are fine to me.

a) is more of a "after long discussions on the first feature, let's go easy for the next few features" one while b) would more be like "let's tackle all difficult topics right at the beginning". Both have their pros and cons - e.g. if you have to report your progress to somebody, probably a) would be good to compensate for the iterations on the first feature.

rustyqt:
Great news, I'll continue with the _olo_ft_ram_sdp_ for the next PR.

PR #314, opened 2026-05-19
Feature/assert entity attribute

obruendl:
unify assert error messages in synthesizable code

Btw: ignore the brancz name. I wanted to use an attribute but had to change the aproach necause vivado does not accept the attribute.

### DIFF
--- a/src/base/vhdl/olo_base_latency_comp.vhd
+++ b/src/base/vhdl/olo_base_latency_comp.vhd
@@ -62,6 +62,9 @@ end entity;
 ---------------------------------------------------------------------------------------------------
 architecture rtl of olo_base_latency_comp is
 
+    -- Constant
+    constant EntityName_c : string := "olo_base_latency_comp[" & AssertsName_g & "]";
+
     -- Entity wide signals
     signal In_Beat  : std_logic;
     signal Out_Beat : std_logic;
@@ -70,7 +73,7 @@ begin
 
     -- *** Assertions ***
     assert compareNoCase(Mode_g, "DYNAMIC") or compareNoCase(Mode_g, "FIXED_CYCLES")
-        report "###ERROR###: olo_base_latency_comp[" & AssertsName_g & "]: Unknown Mode_g - " & Mode_g
+        report errorMessage(EntityName_c, "Unknown Mode_g - " & Mode_g)
         severity error;
 
     -- *** Entity Wide Signals ***
@@ -115,7 +118,7 @@ begin
                     Err_Overrun <= '1';
                     if not AssertsDisable_g then
                         -- synthesis translate_off
-                        report "###WARNING###: olo_base_latency_comp[" & AssertsName_g & "]: Overrun detected in DYNAMIC mode."
+                        report errorMessage(EntityName_c, "Overrun detected in DYNAMIC mode.")
                             severity warning;
                         -- synthesis translate_on
                     end if;
@@ -126,7 +129,7 @@ begin
                     Err_Underrun <= '1';
                     if not AssertsDisable_g then
                         -- synthesis translate_off
-                        report "###WARNING###: olo_base_latency_comp[" & AssertsName_g & "]: Underrun detected in DYNAMIC mode."
+                        report errorMessage(EntityName_c, "Underrun detected in DYNAMIC mode.")
                             severity warning;
                         -- synthesis translate_on
                     end if;
@@ -188,7 +191,7 @@ begin
                     Err_Overrun <= '1';
                     if not AssertsDisable_g then
                         -- synthesis translate_off
-                        report "###WARNING###: olo_base_latency_comp[" & AssertsName_g & "]: Overrun detected in FIXED_CYCLES mode."
+                        report errorMessage(EntityName_c, "Overrun detected in FIXED_CYCLES mode.")
                             severity warning;
                         -- synthesis translate_on
                     end if;
@@ -199,7 +202,7 @@ begin
                     Err_Underrun <= '1';
                     if not AssertsDisable_g then
                         -- synthesis translate_off
-                        report "###WARNING###: olo_base_latency_comp[" & AssertsName_g & "]: Underrun detected in FIXED_CYCLES mode."
+                        report errorMessage(EntityName_c, "Underrun detected in FIXED_CYCLES mode.")
                             severity warning;
                         -- synthesis translate_on
                     end if;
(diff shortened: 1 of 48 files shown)


obruendl:
@LukiLeu 

Please review

src/base/vhdl/olo_base_latency_comp.vhd
```diff
@@ -62,6 +62,9 @@ end entity;
 ---------------------------------------------------------------------------------------------------
 architecture rtl of olo_base_latency_comp is
 
+    -- Constant
+    constant EntityName_c : string := "olo_base_latency_comp[" & AssertsName_g & "]";
```
LukiLeu, line 66: While reviewing, I stumbled upon this block which has the generic `AssertsName_g`. As far as I see, this is the only block that introduces this generic. Is there a reason for that?

src/base/vhdl/olo_base_pkg_logic.vhd
```diff
@@ -292,7 +292,7 @@ package body olo_base_pkg_logic is
         -- Check input width
         -- synthesis translate_off
         assert inp'length mod 8 = 0
-            report "invertByteOrder(): Number of bits must be a multiple of 8"
+            report "olo_base_pkg_logic.invertByteOrder(): Number of bits must be a multiple of 8"
```
LukiLeu, line 296: I suggest using the `errorMessage` function in the packages as well. In my opinion, this has the advantage of ensuring that all messages are formatted consistently.

src/base/vhdl/olo_base_pkg_string.vhd
```diff
@@ -50,6 +50,11 @@ package olo_base_pkg_string is
         a : in string;
         c : in character) return natural;
 
+    -- Print error message
+    function errorMessage (
+        entityName : in string;
```
LukiLeu: If  [https://github.com/open-logic/open-logic/pull/314/changes#r3273955339]( https://github.com/open-logic/open-logic/pull/314/changes#r3273955339) is done, I suggest to change the `entityName` to something more generic like `name`.

tools/inference_test/tools/vivado/synthesize.template
```diff
@@ -12,5 +12,6 @@ add_files {{src_file}}
 
 #Build
 set_property top {{top_entity}} [current_fileset]
+set_property STEPS.SYNTH_DESIGN.ARGS.ASSERT true [get_runs synth_1]
```
LukiLeu, line 15: This should be added to the other templates too. (at least for Quartus, it is disabled by default)

LukiLeu (commented):
Hi @obruendl 

Today, I also did some digging to find a way to retrieve the entity name automatically, but I had no luck. So I guess your approach is probably the most convenient one.

I added some comments; if you agree with these, I can go ahead and make the suggested VHDL changes.

Further, I would also suggest using the `errorMessage` function in the `_vc` components. What do you think?

src/base/vhdl/olo_base_latency_comp.vhd
obruendl: Yes, there is. 

The latency compensation is meant to be placed in parallel to any kind of processing (to compensate the latency for data that bypasses the processing). If a pipeline stage is added to the processing it might happen easily that the latency compensation is not updated - and then assertions are throw. In this case it is very handy for the user to see which latency compensation is off therefore I provide the option to add an identifier.

This is a somewhat expected scenario so I added it there.

I didn't spread the concept to all the other blocks where assertions really are unexpected. But we could discuss if this should be done. What's your take?

src/base/vhdl/olo_base_pkg_logic.vhd
obruendl: Good point. Fixed where possible, placed a comment where not (avoid cyclic dependencies)

src/base/vhdl/olo_base_pkg_string.vhd
obruendl: Done

tools/inference_test/tools/vivado/synthesize.template
obruendl: Do you know the setting for quartus by heart? I thought Vivado is the only tool disabling them by default but I might be wrong.

LukiLeu: For Quartus it is `set_global_assignment -name ENABLE_VHDL_STATIC_ASSERTIONS ON`

src/base/vhdl/olo_base_latency_comp.vhd
LukiLeu: I see, I would not add this to all instances, as it clutters the generics. 

For completeness, a completely different approach in how this might be achieved could be, by using a procedure instead of a function and "bury" the assert in there.

```
   procedure check(
        condition : in boolean;
        message   : in string;
        sev       : in severity_level;
        fallback  : in string) is
        variable reported : boolean := false;
    begin
        reported := false;
        -- synthesis translate_off
        assert condition report fallback'path_name & message severity sev;
        reported := true;
        -- synthesis translate_on

        if not reported then
            assert condition report fallback & message severity sev;
        end if;
    end procedure;
```

During simulation, it would report the path to the constant provided in the fallback.
```
# ** Error: :olo_fix_bin_div_tb:i_dut:entityname_cNumber of iterations must be greater than 0
#    Time: 0 ps  Iteration: 0  Process: /olo_fix_bin_div_tb/i_dut/g_serial/line__123 File: D:/Projekte/open-logic/src/base/vhdl/olo_base_pkg_string.vhd
```

obruendl: I omit this for the moment on my side. Feel free to open a PR. 

The idea seems sane to me - but I currently have a bit of pressure to get forward with things that block contributors.

tools/inference_test/tools/vivado/synthesize.template
obruendl: Probably this is Quartus Pro. For Quartus Standard, I get the following error:

```
[0m------------------------------------------------
ERROR: Illegal assignment: ENABLE_VHDL_STATIC_ASSERTIONS. Specify a legal assignment name.

    while executing
"set_global_assignment -name ENABLE_VHDL_STATIC_ASSERTIONS ON"
    (file "synthesize.tcl" line 13)
------------------------------------------------
```

I therefore omit this here, feel free to investigate and open a PR.

PR #315, opened 2026-05-27
FEATURE: Add RdEna/RdValid to all RAMs

obruendl:
This is a **DRAFT** only.

Not all RAMs are covered yet.

PR #316, opened 2026-06-01
FEATURE: Add olo_ft_ram_sdp (ECC-protected simple dual-port RAM)

rustyqt:
@obruendl Let's continue with the _olo_ft_ram_sdp_:

## Summary

Adds `olo_ft_ram_sdp`, the ECC-protected **simple dual-port** RAM, continuing the fault-tolerant (`ft`) series after the single-port RAM merged in #294. One entity per PR, as agreed.

It wraps `olo_base_ram_sdp` with a SECDED codeword-wide word: data is encoded on write and decoded/corrected on read, transparently to the user. The interface mirrors `olo_ft_ram_sp`:

- Paired `ErrInj_BitFlip` / `ErrInj_Valid` codeword-wide error injection
- `Rd_EccSec` / `Rd_EccDed` status flags, time-aligned with `Rd_Data`
- `Rd_Valid` pulse = read-enable delayed by `RamRdLatency_g + EccPipeline_g`
- Independent read/write clocks via `IsAsync_g` (`Rd_Clk` / `Rd_Rst`)

Built directly on the official base-RAM `RdEna` / `RdValid` interface from #315 (the reset is always propagated into the inner base RAM so the read-valid pipeline is reset correctly).

## Contents

- `src/ft/vhdl/olo_ft_ram_sdp.vhd` + testbench
- `doc/ft/olo_ft_ram_sdp.md` with architecture diagram
- Sim test configs, inference-test config, `EntityList.md` entry, regenerated `compile_order.txt`

The testbench includes a `ResetInFlight` case: a reset asserted while reads are in flight (`Rst` in sync mode, `Rd_Rst` in async mode) must squash the read-valid pipeline even under continuous `Rd_Ena`, must not leak a stale valid after release, and the RAM must recover. `EccPipeline_g` is swept 0..2 to match `olo_ft_ram_sp`.

## Verification

- GHDL regression green
- `vsg --all_phases` clean (source + testbench)
- markdownlint clean (docs)
- Libero inference: maps to 1 block RAM (no register fallback)


obruendl (approved):
All good - no comments

PR #317, opened 2026-06-02
Feature/fix nvc coverage commands

tasgomes:
This PR integrates NVC coverage into the newly updated `run.py` that handles arguments as VUnit arguments.

I think this PR makes the branch `feature/nvc-coverage` unnecessary, which probably should be deleted. Please review my assumption before doing so.


obruendl:
@tasgomes 

The _feature/nvc-coverage_ branch has quite a few other changes. 

It also adds interpretation of the coverage results to the _AnalyzeCoverage.py_ and it also makes sure the ncdb files all have unique names (because depending on the VUnit version they end up all in the same folder) and has an updated gitignore.

Is there anything present in your branch that is missing in _feature/nvc-coverage_? I could not identify anything - I think all your changes are present in _feature/nvc-coverage_ as well.

Let me know if I missed anything.

tasgomes:
I have merged _feature/nvc-coverage_ into this branch. There was a conflict in _AnalyzeCoverage.py_, I hope it resolved the correct way.

> it also makes sure the ncdb files all have unique names (because depending on the VUnit version they end up all in the same folder) 

I have removed the renaming of ncdb and I upgraded VUnit version accordingly so that this renaming is not necessary anymore and does not interfere with newest VUnit versions.

_feature/nvc-coverage_ still has the old `run.py` arguments, so this branch has now solved that merge conflict.

sim/AnalyzeCoverage.py
```diff
@@ -23,20 +23,18 @@
     default=0.0,
     help="Minimum coverage percentage required (default: 0.0)."
 )
+parser.add_argument(
+    "--simulator",
+    choices=["modelsim", "nvc"],
+    default="modelsim",
+    help="Simulator to use for coverage analysis (default: nvc)."
```
Gaweringo, line 30: Default in help string and `default` don't match.

Sorry for plopping in here unannounced.

PR #319, opened 2026-06-04
FEATURE: Add olo_ft_ram_tdp (ECC-protected true dual-port RAM)

rustyqt:
## Summary

Adds `olo_ft_ram_tdp`, the ECC-protected **true dual-port** RAM, continuing the fault-tolerant (`ft`) series after the single-port (#294) and simple dual-port (#316) RAMs. One entity per PR, as agreed.

It wraps `olo_base_ram_tdp` with a SECDED codeword-wide word. Each port (A and B) has its **own** ECC encoder and decoder, so the two ports are protected independently and run on their own clocks. The interface follows `olo_ft_ram_sp` / `olo_ft_ram_sdp`, per port with `A_`/`B_` prefixes:

- Paired `A_ErrInj_BitFlip` / `A_ErrInj_Valid` (and `B_`) codeword-wide error injection
- `A_RdEccSec` / `A_RdEccDed` status flags, time-aligned with `A_RdData`
- `A_RdValid` pulse = read-enable delayed by `RamRdLatency_g + EccPipeline_g`

Built on the official base-RAM `RdEna` / `RdValid` interface from #315. Note the TDP specific behavior: the base true-dual-port RAM **always reads**, so `A_RdEna` / `B_RdEna` only gate the corresponding `RdValid` (they do not gate the read itself).

## Contents

- `src/ft/vhdl/olo_ft_ram_tdp.vhd` + testbench
- `doc/ft/olo_ft_ram_tdp.md` with architecture diagram
- Sim test configs, inference-test config, `EntityList.md` entry, regenerated `compile_order.txt`

The testbench covers both ports, cross-port access, SEC/DED on each port, all-codeword-bit SEC sweeps and sampled DED pairs. `EccPipeline_g` is swept 0..2 to match `olo_ft_ram_sp` / `olo_ft_ram_sdp`.

## Constraint

True dual-port RAM is **not** supported when compiling with Yosys for Cologne Chip FPGAs (inherited from `olo_base_ram_tdp`); documented in the entity doc.

## Verification

- GHDL regression green
- `vsg --all_phases` clean (source + testbench)
- markdownlint clean (docs)


obruendl:
@rustyqt 

I checked synthesis results for the TDP RAM and fixed the inference test yaml to work fine for all tools.

At this instance I noticed that the olo_ft area was missing in many import sources scripts. I added it there.

PR #320, opened 2026-06-06
FEATURE: Add olo_ft_ram_sp_scrub and olo_ft_ram_sdp_scrub (ECC RAM scrubbers)

rustyqt:
## Summary

Adds opportunistic background memory scrubbers for the ECC-protected single-port and simple
dual-port RAMs:

- `olo_ft_ram_sp_scrub` - wraps `olo_ft_ram_sp`
- `olo_ft_ram_sdp_scrub` - wraps `olo_ft_ram_sdp`
- `olo_ft_private_scrubber` - shared, reusable scrubber engine instantiated by both wrappers

Both scrub wrappers reuse the same engine, so they are submitted together in one PR.

## The scrubber engine

`olo_ft_private_scrubber` owns the user/scrubber arbitration, the SEC-only writeback FSM, and the
user read-valid mask. It walks the address space and rewrites a word whenever the wrapped RAM
reports a correctable single-bit error, opportunistically using only cycles where the user is idle
(the user always wins; user accesses are never stalled). The wrappers reduce to RAM + engine
wiring: `sp_scrub` collapses the muxed write/read channels onto the single RAM port, `sdp_scrub`
maps them 1:1 onto the dedicated write/read ports.

## Scope

There is intentionally **no `tdp_scrub`**, and the scrub variants are **sync-only**: a
true-dual-port or generic-async scrubber would require cross-clock-domain handling, which is out
of scope for now.

## Included

Sources, testbenches, per-entity documentation with architecture diagrams, and the
`EntityList.md` / `ft.yml` / `olo_ft.py` / `compile_order.txt` integration. The `ft.yml` blocks
follow the `in_reduce`/`out_reduce` convention so the inference test runs across all tools.

This PR also wires the `ft` area into FuseSoC packaging (`src/ft/olo_ft_dev.core` +
`UpdateCoreFiles.py` entries; the stable core is left to the release-time regeneration) and adds
the `ft` inference-test step to the CI synthesis workflow.

## Verification

- Full GHDL regression: all passed (5846/5846), including the sp_scrub / sdp_scrub sweeps
  (RamBehavior x RamRdLatency x EccPipeline 0..2 x Width).
- The user-write-wins arbitration is mutation-verified: removing the write-side inhibit from the
  scrubber engine fails the new `UserWriteToInFlightScrubAddr` test in every
  TotalReadLatency >= 2 configuration.
- `vsg` lint clean on all new sources and testbenches.
- markdownlint clean on all docs.

PR #335, opened 2026-08-06
FEATURE: Add olo_ft_fifo_sync and olo_ft_fifo_packet (ECC-protected synchronous FIFOs)

rustyqt:
## What

Next installment of the fault-tolerant series (roadmap in the [#320 comment](https://github.com/open-logic/open-logic/pull/320#issuecomment-5203028242)): the two synchronous FIFOs.

- `olo_ft_fifo_sync`: ECC-protected synchronous FIFO, interface matches `olo_base_fifo_sync`
- `olo_ft_fifo_packet`: ECC-protected packet FIFO, interface matches `olo_base_fifo_packet` (deliberate deviations below)
- Shared TB procedures `ftPushBeat` / `ftExpectBeat` / `ftExpectedBeat` added to `olo_test_ft_pkg` (AXI-Stream push/expect with error injection; reused by the upcoming ft PRs)

## Architecture

Both entities are exactly the codec sandwich the ECC encoder / decoder were built for: `olo_ft_ecc_encode` in front, the unchanged base FIFO at codeword width in the middle, `olo_ft_ecc_decode` behind it. No base code is touched. The locked ft API applies as in the merged entities: latched `ErrInj_BitFlip` / `ErrInj_Valid` injection pair, `Out_EccSec` / `Out_EccDed` status flags, `EccPipeline_g` 0..2.

Both docs include an architecture diagram in the style of the ft RAM docs (editable drawio PNGs under `doc/ft/fifo/`).

For the packet FIFO, the main RAM stores pure codewords. The small packet-boundary FIFO (packet end addresses, from which `Out_Last` / `Out_Size` / `Out_Next` derive) is not part of the data codeword; therefore `SmallRamStyle_g` defaults to `"registers"` so that state lands in flip-flops coverable by vendor TMR. Details in the "Fault-Tolerant Storage of Packet Boundaries" section of the doc.

## Deliberate deviations on olo_ft_fifo_packet

- `FeatureSet_g = "DROP_ONLY"` is rejected by an elaboration assert. In the base implementation, DROP_ONLY stores `In_Last` of every word inside the main RAM, outside the ECC parity; a bit flip there would silently split or merge packets. FULL and DROP_SKIP_ONLY keep the main RAM a pure codeword store and are fully supported.
- `SmallRamStyle_g` defaults to `"registers"` (base: `"auto"`), see above.
- `MaxPacketSize_g` and `Optimization_g` are not exposed; the wrapper uses the base defaults (`-1` / `"SPEED"`). Happy to expose them if you prefer strict generic parity, let me know.

Also documented: with `EccPipeline_g > 0`, `Out_Next` / `Out_Repeat` are sampled on the internal FIFO handshake, which runs ahead of the observed output by the decode pipeline; the doc describes the safe usage.

## Verification

- Full GHDL regression on this branch: 6048 / 6048 passed
- Test benches cover both feature sets, `EccPipeline_g` 0..2, drop / repeat semantics, reset in flight, and error-injection sweeps (SEC on every bit position, DED pairs, latched-injection timing)
- vsg 3.27: zero violations on all changed VHDL files
- markdownlint: clean
- `ft.yml` carries one representative config per entity for the CI synthesis workflow; `compile_order.txt` and `olo_ft_dev.core` regenerated


doc/EntityList.md
```diff
@@ -307,6 +308,13 @@ common constraints) are described once in
 | [olo_ft_ram_sdp_scrub](./ft/olo_ft_ram_sdp_scrub.md) | ECC-protected simple dual-port RAM with an opportunistic background memory scrubber |
 | [olo_ft_ram_tdp](./ft/olo_ft_ram_tdp.md) | ECC-protected true dual-port RAM                             |
 
+### FIFO Implementations (olo_ft_fifo_\<...\>)
+
+| Entity                                               | Description                                                  |
+| ---------------------------------------------------- | ------------------------------------------------------------ |
+| [olo_ft_fifo_sync](./ft/olo_ft_fifo_sync.md)         | ECC-protected synchronous FIFO (single clock)               |
+| [olo_ft_fifo_packet](./ft/olo_ft_fifo_packet.md)     | ECC-protected packet FIFO (store and forward) with the ability to drop packets on the write side and skip or repeat packets on the read side |
+
```
obruendl, line 317: IMO it would make sense to to mention why t here is no _fifo_async_. IIRC the conclusion was that the grey code CDC is unsafe in fault tolerant environments. Correct? I'd mention this here - just to prevent questions.

doc/EntityList.md
```diff
@@ -307,6 +308,13 @@ common constraints) are described once in
 | [olo_ft_ram_sdp_scrub](./ft/olo_ft_ram_sdp_scrub.md) | ECC-protected simple dual-port RAM with an opportunistic background memory scrubber |
 | [olo_ft_ram_tdp](./ft/olo_ft_ram_tdp.md) | ECC-protected true dual-port RAM                             |
 
+### FIFO Implementations (olo_ft_fifo_\<...\>)
+
+| Entity                                               | Description                                                  |
```
obruendl, line 313: I'd mention the non-ECC versions in the Description. Saying something like "analog to ..."

doc/ft/olo_ft_fifo_sync.md
```diff
@@ -0,0 +1,123 @@
...
+| Name       | In/Out | Length                  | Default | Description                                                  |
+| :--------- | :----- | :---------------------- | ------- | :----------------------------------------------------------- |
+| Out_Data   | out    | _Width_g_               | N/A     | Output data (corrected if a single-bit error was detected)   |
+| Out_Valid  | out    | 1                       | N/A     | Output valid (AXI-S handshaking)                             |
+| Out_Ready  | in     | 1                       | '1'     | Output ready (AXI-S handshaking)                             |
+| Out_Level  | out    | _ceil(log2(Depth_g+1))_ | N/A     | Output-side fill level                                       |
+| Out_EccSec | out    | 1                       | N/A     | Single error corrected flag. Time-aligned with _Out_Data_.   |
+| Out_EccDed | out    | 1                       | N/A     | Double error detected flag. Read data is unreliable. Time-aligned with _Out_Data_. |
+
+### Status
```
obruendl, line 67: Do these signals make sense in the FT implementation?

When EccPipeline_g is > 0, then these signals are delayed, which IMO is confusing. Since Full/Empty are womewhat redundant to _Out_Valid_ and _In_Ready_ and the "almost..." functionality is not a very basic need, would it make sense to better omit those signals than to expose them in a delayed form which could cause confusion?

What's your take?

In case we remove them, of course we can also drop all the related Generics

doc/ft/olo_ft_fifo_sync.md
```diff
@@ -0,0 +1,123 @@
...
+| :-------------- | :-------- | ------- | :----------------------------------------------------------- |
+| Width_g         | positive  | -       | Number of data bits per FIFO entry. The internal FIFO is wider to accommodate ECC parity bits. |
+| Depth_g         | positive  | -       | Number of entries the FIFO has                               |
+| AlmFullOn_g     | boolean   | false   | Enable almost-full flag                                      |
+| AlmFullLevel_g  | natural   | 0       | Almost-full threshold level                                  |
+| AlmEmptyOn_g    | boolean   | false   | Enable almost-empty flag                                     |
+| AlmEmptyLevel_g | natural   | 0       | Almost-empty threshold level                                 |
+| RamStyle_g      | string    | "auto"  | Controls the RAM implementation resource                     |
+| RamBehavior_g   | string    | "RBW"   | Controls the RAM behavior. "RBW" or "WBR"                    |
+| ReadyRstState_g | std_logic | '1'     | Value of _In_Ready_ during reset                             |
```
obruendl, line 35: I think this should be removed, the Fifo "ReadyRstState_g" should be fixed to '1' and this behavior should be documented.

Reason: The _olo_base_pl_stage_ always asserts In_Ready='1' on reset and this is not configurable. When we have the pipeline stage enabled in the encode block, then the behavior of the pipeline stage is exposed on the input.

Alternatively if you want to keep the option, we have to make the _olo_base_pl_stage_ ready behavior on reset configurable.

doc/ft/fifo/olo_ft_fifo_sync_arch.drawio.png
obruendl, line 1: See comment mentioned in the doc: I am not sure if we want to keep the Empty/Full/AlmEmpty/AlmFull signals.

doc/ft/olo_ft_fifo_sync.md
```diff
@@ -0,0 +1,123 @@
...
+| Rst  | in     | 1      | -       | Reset (high-active, synchronous to _Clk_). Empties the FIFO and clears the internal error-injection latch and the codec pipeline. |
+
+### Input Data
+
+| Name     | In/Out | Length                  | Default | Description                     |
+| :------- | :----- | :---------------------- | ------- | :------------------------------ |
+| In_Data  | in     | _Width_g_               | -       | Input data                      |
+| In_Valid | in     | 1                       | '1'     | Input valid (AXI-S handshaking) |
+| In_Ready | out    | 1                       | N/A     | Input ready (AXI-S handshaking) |
+| In_Level | out    | _ceil(log2(Depth_g+1))_ | N/A     | Input-side fill level           |
```
obruendl, line 54: The level I'd potentially keep (often useful status information). However, we have the delay issue here as well.

I see the following options:
* Document it is "EccPipeline_g" cycles late on the read side.
* Add a separate counter that increments when a sample enters the ECC encoder and decrements when a sample leaves the ECC decoder. This needs additional bits to cover the samples that are in the ECC decoder.
* Fix t he decoder pipeline stages to 0 (at the cost of lower clock)
* Omit it (at the cost of less information)

Comment applies to Input and Output level.

What's your take?

doc/ft/olo_ft_fifo_sync.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Detailed Description
+
+### Architecture
+
+![olo_ft_fifo_sync architecture](./fifo/olo_ft_fifo_sync_arch.drawio.png)
+
+The FIFO is a straight pipeline of three Open Logic entities:
+
+1. [olo_ft_ecc_encode](./olo_ft_ecc_encode.md) encodes each accepted input beat into a SECDED codeword
```
obruendl, line 96: Mention that this is always combinatorial (zero cycles delay)

doc/ft/olo_ft_fifo_sync.md
```diff
@@ -0,0 +1,123 @@
...
+   (combinational, the AXI-S handshake passes through).
+2. [olo_base_fifo_sync](../base/olo_base_fifo_sync.md) stores the codeword (entity configured with a
```
obruendl, line 98: Update to reflect decisions on status signals.

doc/ft/olo_ft_fifo_sync.md
```diff
@@ -0,0 +1,123 @@
...
+   codeword-wide word). All levels and status flags come directly from the base FIFO.
+3. [olo_ft_ecc_decode](./olo_ft_ecc_decode.md) decodes and corrects each beat on the read side and drives
+   _Out_EccSec_ / _Out_EccDed_ time-aligned with _Out_Data_. `EccPipeline_g` inserts register stages on the
+   decode datapath (see the [codec documentation](./olo_ft_ecc_decode.md) for the stage placement).
+
+The codeword is protected end-to-end while it is inside the FIFO: encoding happens before, and decoding after,
+all storage elements.
+
+See [olo_base_fifo_sync](../base/olo_base_fifo_sync.md) for detailed FIFO behavior.
+
```
obruendl, line 108: I'd add a note that the FIFO does NOT come with a scrubber because FIFOs do not store data permanently and hence the risk for error accumulation is not given.

test/ft/olo_ft_fifo_sync/olo_ft_fifo_sync_tb.vhd
```diff
@@ -0,0 +1,343 @@
...
+-- vunit: run_all_in_same_sim
+entity olo_ft_fifo_sync_tb is
+    generic (
+        runner_cfg    : string;
+        Width_g       : positive range 5 to 128 := 32;
+        EccPipeline_g : natural range 0 to 2    := 0
+    );
+end entity;
+
+architecture sim of olo_ft_fifo_sync_tb is
```
obruendl, line 38: Depending on decisions about the status signals, other cases have to be added (in case we decide for a separate level counter, this must be tested).

test/ft/olo_ft_fifo_sync/olo_ft_fifo_sync_tb.vhd
```diff
@@ -0,0 +1,343 @@
...
+                Flip_v := (others => '0');
+                ftPushBeat(net, AxisMaster_c, Clk, In_ErrInj_BitFlip, In_ErrInj_Valid, toUslv(16#01#, Width_g), Flip_v);
+                ftPushBeat(net, AxisMaster_c, Clk, In_ErrInj_BitFlip, In_ErrInj_Valid, toUslv(16#02#, Width_g), setBits(0, CodewordWidth_c));
+                ftPushBeat(net, AxisMaster_c, Clk, In_ErrInj_BitFlip, In_ErrInj_Valid, toUslv(16#03#, Width_g), Flip_v);
+                ftExpectBeat(net, AxisSlave_c, toUslv(16#01#, Width_g), (Flip_v'range => '0'),         "Mixed[0] clean");
+                ftExpectBeat(net, AxisSlave_c, toUslv(16#02#, Width_g), setBits(0, CodewordWidth_c),  "Mixed[1] Sec");
+                ftExpectBeat(net, AxisSlave_c, toUslv(16#03#, Width_g), (Flip_v'range => '0'),         "Mixed[2] clean");
+
+            ---------------------------------------------------------------------------------------
+            elsif run("FullEmpty") then
```
obruendl, line 146: The Full and Empty flags are not checked here. They should.

Yes, they come from the underlying FIFO - but this could be changed when the entity is modified and the TB shall check the complete feature set, so we have this covered.

test/ft/olo_ft_fifo_sync/olo_ft_fifo_sync_tb.vhd
```diff
@@ -0,0 +1,343 @@
...
+            In_ErrInj_BitFlip <= (others => '0');
+            In_ErrInj_Valid   <= '0';
+            wait until rising_edge(Clk);
+            Rst               <= '1';
+            wait for 200 ns;
+            wait until rising_edge(Clk);
+            Rst               <= '0';
+            wait until rising_edge(Clk);
+
+            ---------------------------------------------------------------------------------------
```
obruendl, line 109: The levels do not seem to be tested (or did I overlook them)?

They come from the underlying FIFO, but we should still test them. Could be that we change this decision (it's even a consideration in another thread) - and the TB should not omit tests for things just because they come from underlying components (this can change in future)

obruendl (commented):
I reviewed the Sync FIFO only for the moment. Let's do that first and then switch to the packet FIFO.

doc/EntityList.md
rustyqt: The async FIFO is a bit more involved and I would defer it to a separate PR. Protecting
olo_base_fifo_async is not just a codec sandwich: the Gray pointer / reset CDC paths need
fault-tolerant synchronizers, so the async FIFO brings a small set of olo_ft_cc_* entities plus
an attributes package with it, and it needs a minor restructure of olo_base_fifo_async
(extracting the core logic into a private entity so the ft variant can compose it). So, I probably
would even split it in two PRs, one for adding the olo_ft_cc_* entities and then one for the async fifo.

rustyqt: We can add a small note that clarifies that the olo_ft_async fifo will be added in a later PR as it requires additional CDC primitives.

doc/EntityList.md
rustyqt: Agreed

doc/ft/olo_ft_fifo_sync.md
rustyqt: Agreed to drop full and empty could be dropped, but I would leave `AlmFull ` and `AlmEmpty ` and describe their behavior in the documentation as they sometimes become useful flow control indicators in scenarios where you are trying to integrate building blocks / IPs that are not backpressurable.

doc/ft/olo_ft_fifo_sync.md
rustyqt: The `olo_ft_ecc_encode.Pipeline_g` is fixed to 0 in the `olo_ft_fifo_sync`. This means the `olo_base_pl_stage` is in pass through mode. --> `In_Ready` is the `olo_base_fifo_sync.In_Ready` output and therefore preserves the `olo_base_fifo_sync` behavior.

doc/ft/olo_ft_fifo_sync.md
rustyqt: There is also a good example in open-logic that uses the `AlmEmpty` flag:
https://github.com/open-logic/open-logic/blob/main/src/base/vhdl/olo_base_flowctrl_handler.vhd

So I would argue they should be preserved. I would even argue to keep the `Full` and `Empty` to preserve the drop-in replacability.

doc/ft/olo_ft_fifo_sync.md
rustyqt: Agreed

doc/ft/olo_ft_fifo_sync.md
rustyqt: My main goal is that the `ft` variants can be used as drop-in replacements to their `base` counterparts. That's why I would like to preserve the functionality as much as possible even if that means to add more complicated fill level logic to achieve this goal. Limitations that affect clock speed I also would like to avoid. So, I tend towards something similar to your second option:

- Adding additional `Fifo_In_Level` and `Fifo_Out_Level` signals that are calculated by adding `olo_base_fifo_sync.In_Level` and `olo_base_fifo_sync.Out_Level` to the counted samples in the decoder stages (encoder is pass-through by design `olo_ft_ecc_encode.Pipeline_g` = 0)
- `Fifo_Out_Level` and `Fifo_In_Level` should be `std_logic_vector(log2ceil(Depth_g + 2*EccPipeline_g + 1) - 1 downto 0);`  to fit the additional samples in the decoder stages and routed to the top level ports `In_Level` and `Out_Level`
- `Full` and `AlmFull` will be dervied from `Fifo_In_Level`
- `Empty` and `AlmEmpty` will be derived from `Fifo_Out_Level`
- `olo_base_fifo_sync.AlmFullOn_g` and `olo_base_fifo_sync.AlmEmptyOn_g` will be tied to false, to avoid paying the comparison logic cost twice.

doc/ft/fifo/olo_ft_fifo_sync_arch.drawio.png
rustyqt: Tend to keep them in order to preserve the `ft` variant drop-in replacement capabilities as much as possible.

doc/ft/olo_ft_fifo_sync.md
rustyqt: Yes, agreed.

doc/ft/olo_ft_fifo_sync.md
rustyqt: Yes, agreed, needs updating.

test/ft/olo_ft_fifo_sync/olo_ft_fifo_sync_tb.vhd
rustyqt: Agreed.

rustyqt:
@obruendl

Since the pipelining of the ECC decode came up in several threads, I measured the maximum frequency
of both FIFOs over `EccPipeline_g`. Numbers below, they answer the "at the cost of lower clock"
question in the fill-level thread with data.

**Setup:** Libero 2025.2, MPFS250T_ES / FCVG484, STD speed, EXT range. `Width_g=32`, `Depth_g=256`.
The DUT is wrapped in an `olo_base_pl_stage` (`Stages_g=1`, `UseReady_g=true`) on the input and on
the output side, so the combinational encode and decode paths are timed register to register. The
output stage carries the sidebands (`Out_EccSec`/`Out_EccDed`, plus `Out_Last`/`Out_Size` for the
packet FIFO) together with the data. All status ports are connected so the level and flag logic is
not optimized away.

### olo_ft_fifo_sync

| EccPipeline_g | Fmax (MHz) | LUTs | SLEs | Block RAM |
| ------------- | ---------- | ---- | ---- | --------- |
| 0             | 153.2      | 387  | 361  | 1         |
| 1             | 171.5      | 376  | 432  | 1         |
| 2             | 226.6      | 423  | 513  | 1         |

### olo_ft_fifo_packet

| EccPipeline_g | Fmax (MHz) | LUTs | SLEs | Block RAM |
| ------------- | ---------- | ---- | ---- | --------- |
| 0             | 154.9      | 694  | 610  | 1         |
| 1             | 167.0      | 696  | 703  | 1         |
| 2             | 176.1      | 758  | 792  | 1         |

### Observations

The critical path starts at the block RAM output in all six runs. The read path through the ECC
decode dominates, the write path through the encoder is never critical.

For `olo_ft_fifo_sync` the scaling is real. At `EccPipeline_g=0` the path runs from the RAM through
the complete syndrome and correction logic into the harness output register. At 1 it terminates in
the decoder's own output register (`i_dec/i_pl2`), same logic depth, +12%. At 2 it terminates in
`i_dec/i_pl1`, the stage that `olo_ft_ecc_decode` inserts between syndrome and correction, which is
where the gain comes from: +32% over `EccPipeline_g=1` and +48% over 0.

For `olo_ft_fifo_packet` the second stage does nothing for timing. Its decoder is instantiated with
`Pipeline_g => 0` and `EccPipeline_g` drives the `olo_base_pl_stage` that carries data plus the
Last/Size/Sec/Ded sidebands, so both stages sit in series after the decode and the decode itself is
never split. The timing report confirms it: at `EccPipeline_g=2` the path still ends at stage 0 of
that pipeline, exactly as at 1. The remaining 5% is placement noise, paid for with a cycle of
latency, 62 LUTs and 89 SLEs.

Side by side, the two entities are equivalent up to one stage and then diverge:

| EccPipeline_g | sync | packet |
| ------------- | ---- | ------ |
| 0             | 153.2 | 154.9 |
| 1             | 171.5 | 167.0 |
| 2             | 226.6 | 176.1 |

### Consequences

For the fill-level discussion: fixing the decoder pipeline to 0 costs about a third of the
achievable clock on the sync FIFO (153 vs 227 MHz), so I would rather solve the level question in
logic than by removing the pipelining option.

For `olo_ft_fifo_packet` I see two ways forward and would like your opinion. Either limit
`EccPipeline_g` to 0..1 and document that the packet FIFO does not scale beyond one stage, or rework
the packet FIFO so it gets the same scaling as the sync FIFO by pipelining the decoder and carrying
the sidebands alongside it. The second option is the better result but it moves the alignment of
data, Last and Size from a structural property (one pipeline, one handshake) to something that
depends on additional control logic staying in step, although it could be simply assumed that input and output for these parallel pipeline stages are always identical.

doc/ft/olo_ft_fifo_sync.md
rustyqt: One refinement to the sketch above, after going through the base FIFO in detail.

In `olo_base_fifo_sync`, `Full` and `In_Ready` are the same condition: both are `WrLevel = Depth_g`
(and likewise `Empty` and `Out_Valid` are both derived from `RdLevel > 0`). If `In_Level` carries
`WrLevel` plus the decoder count, then no threshold reproduces that equivalence:

- with the threshold at `Depth_g + 2*EccPipeline_g`, `Full` only asserts when the FIFO is full *and*
  the decoder happens to be completely full as well. So whenever the reader is draining, `Full`
  stays low although the FIFO is rejecting writes.
- with the threshold at `Depth_g`, `Full` asserts while the FIFO still accepts data.

The reason is that the write side has no pipeline in front of it. `olo_ft_ecc_encode` is
instantiated with `Pipeline_g => 0`, so a beat accepted at `In_Valid` / `In_Ready` reaches the FIFO
in the same clock cycle and `WrLevel` is already exact. The decoder sits behind the FIFO, so only
the read side is affected by it.

I would therefore keep your second option, but scope the correction to the read side:

| Port | Source | Result |
| :--- | :----- | :----- |
| `In_Level`, `Full`, `AlmFull` | base FIFO, untouched | bit-identical to `olo_base_fifo_sync` |
| `Out_Level` | base `Out_Level` + decoder count, widened to `log2ceil(Depth_g + 2*EccPipeline_g + 1)` | exact entity content |
| `Empty` | `Out_Level = 0` | exact, safe as a drain condition |
| `AlmEmpty` | `Out_Level <= AlmEmptyLevel_g` | exact |

rustyqt:
@obruendl

Pushed `ffb5f13`, which answers the review round on the synchronous FIFO. The main points:

**Read-side status corrected.** With `EccPipeline_g > 0` the decoder buffers up to
`2 * EccPipeline_g` beats, so status taken from the internal FIFO alone did not describe the entity:
`Empty` asserted while `Out_Valid` was still delivering data. `Out_Level`, `Empty` and `AlmEmpty` are
now produced by `olo_private_ft_fifo_status`, a small two-process entity in the same file that counts
what the FIFO and the decoder still have to deliver. `Out_Level` is widened to
`log2ceil(Depth_g + 2*EccPipeline_g + 1)`, and `AlmEmptyOn_g` is tied off in the internal FIFO so the
comparison is not paid for twice.

**Write side untouched.** `In_Level`, `Full` and `AlmFull` stay bit-identical to
`olo_base_fifo_sync`. The encoder is combinational, so a beat reaches the FIFO in the cycle it is
accepted, `Full` remains the exact inverse of `In_Ready`, and `ReadyRstState_g` reaches the port
unmodified. That is also why I kept `ReadyRstState_g`: the encoder is instantiated with
`Pipeline_g => 0`, so `olo_base_pl_stage` is in pass-through mode and its reset behavior is never
exposed.

**All four flags kept**, as discussed, with the semantics documented rather than removed.

**Test bench.** Two monitors now run in every test case: `Empty` never high while `Out_Valid` offers
a beat, and `Full` always the inverse of `In_Ready`. Both fail against the previous code. Coverage is
brought to parity with `olo_base_fifo_sync_tb`: `ReadEmptyFifo`, `WriteFullFifo`,
`AlmostFlagsThresholds`, `StatusAndLevels` and `ReadyRstState`, plus the `Depth_g`, `AlmFullOn_g`,
`AlmEmptyOn_g`, `RamBehavior_g` and `ReadyRstState_g` sweeps. `Depth_g = 31` is in the sweep because that is the case in
which the decoder beats widen `Out_Level` beyond the base FIFO width.

**Documentation** covers the status semantics, the combinational encoder and the reason the FIFOs
ship without a scrubber. The EntityList now names the non-ECC counterparts and states that the
asynchronous FIFO follows once the ft clock-crossing primitives are in. The architecture diagram was
redrawn with the status entity.

Verified: full GHDL regression 6243/6243, vsg 3.27 clean, markdownlint clean.

One open point from the level thread: widening `Out_Level` costs width compatibility for a drop-in
swap. I prefer a loud compile error over a silently wrong value, but if you would rather have both
levels the same width I can widen `In_Level` as well and zero-extend it.

rustyqt:
I was just checking the `olo_ft_fifo_packet` for the status signal alignment and more important the Out_Next and Out_Repeat inputs are all breaking with a decoder pipeline. For the olo_ft_fifo_packet I have settled to force `EccPipeline_g` to 0 for the decoder. 

Although the fmax scaling is temping, realistically, 150 Mhz is more than enough for a Polarfire design. If we apply the same reasoning to the `olo_ft_fifo_sync` it would be the correct decision to simplify this entity as well by forcing the decoder `EccPipeline_g` to 0. In this case I will revert the changes from the last commit and implement the simplified version instead.

rustyqt:
@obruendl

Pushed `0d66d03`, which **supersedes `ffb5f13`**. Please review the current state rather than that
commit, the design changed after I looked at the packet FIFO.

**Why:** I went through `olo_ft_fifo_packet` to apply the same status-alignment reasoning and found
that the read-side *control* does not survive an output pipeline at all. `Out_Next` and `Out_Repeat`
are sampled on the internal FIFO handshake, which runs ahead of the beats the user observes, so a
reader deciding from observed data skips or repeats the wrong beats and the error is silent.
`PacketLevel` has the same issue, and the repeat state needed to correct it is internal to the base
FIFO. Unlike the level skew in the sync FIFO, this is not correctable from the wrapper.

**Decision:** Both FIFOs now instantiate encoder and decoder with `Pipeline_g = 0` and offer no
`EccPipeline_g` generic. The measured cost is modest: the packet FIFO plateaus at 176 MHz even at
`EccPipeline_g = 2` (see the Fmax numbers I posted earlier), and the sync FIFO reaches 153 MHz
combinational, which is good enough. 

**What this removes:** `EccPipeline_g` on both entities, `olo_private_ft_fifo_status`, the widened
`Out_Level`, the sideband bundle in the packet FIFO, and the caveat that `Out_Next` / `Out_Repeat`
act on a point the user cannot see. Both entities are now three instantiations and nothing else, all
levels and status flags come straight from the internal FIFO and are bit-identical to the base
entities. That also closes the open question from the level thread: `Out_Level` is back to the base
width, so drop-in replacement is exact again.

**Test benches kept and extended:** The monitors and the parity cases against
`olo_base_fifo_sync_tb` from `ffb5f13` remain. The packet test bench gained the families it was
missing against `olo_base_fifo_packet_tb`: `NextPacket` (`Out_Next` had no coverage at all before),
`PacketLevelAndFreeWords`, `Size1Packets`, `Wraparound` and `ResetState`. I did not mirror the base
suite's full drop/repeat/next permutation matrix, since those exercise the base FIFO's packet
machinery which the wrapper does not touch.

Verified: full GHDL regression 6199/6199, vsg 3.27 clean, markdownlint clean. Both architecture
diagrams redrawn.

test/ft/olo_ft_fifo_sync/olo_ft_fifo_sync_tb.vhd
rustyqt: Agreed, done. Both flags are now checked in a directed `StatusAndLevels` case at reset, mid fill, full and drained, and by two monitors that run in *every* test case: `Empty` is never high while `Out_Valid` offers a beat, and `Full` is always the inverse of `In_Ready`. `ReadEmptyFifo` and `WriteFullFifo` cover the two boundaries directly.

test/ft/olo_ft_fifo_sync/olo_ft_fifo_sync_tb.vhd
rustyqt: Agreed, done. `In_Level` and `Out_Level` are checked in `StatusAndLevels` at reset, mid fill, full and drained, and in `ReadEmptyFifo` after a read attempt on an empty FIFO. The test bench now also sweeps `Depth_g` (31, 53, 128), so the level widths are exercised at odd and power-of-two depths.

doc/ft/olo_ft_fifo_sync.md
rustyqt: Superseded by `0d66d03`. Reviewing `olo_ft_fifo_packet` showed that an output pipeline also breaks `Out_Next` / `Out_Repeat` there, and that part is not correctable from the wrapper. Both FIFOs therefore instantiate the codecs with `Pipeline_g = 0` and drop `EccPipeline_g` entirely, so no level correction is needed at all: every level and flag comes straight from the internal FIFO and is bit-identical to the base entity. `Out_Level` keeps the base width, which also settles the drop-in width question raised above.

obruendl:
I am currently on holiday in south africa and I have very limited network access.

I'll review your changes when I am back.

doc/ft/olo_ft_fifo_packet.md
```diff
@@ -0,0 +1,164 @@
...
+Both codecs are instantiated with `Pipeline_g = 0`. The datapath to and from the internal FIFO is
+therefore combinational, and the _ft_ entity behaves exactly like its
+[olo_base_fifo_packet](../base/olo_base_fifo_packet.md) counterpart.
+
+The ECC decode lies between the RAM output and the output ports and is the critical path of the entity.
+If it limits the achievable clock frequency, add an
+[olo_base_pl_stage](../base/olo_base_pl_stage.md) on the output side in the surrounding design.
+Register _Out_Data_, _Out_Last_, _Out_Size_, _Out_EccSec_ and _Out_EccDed_ in the same stage to keep them
+aligned. Note that _Out_Next_ and _Out_Repeat_ then refer to the beat inside that stage, not to the one
+observed behind it.
```
obruendl: If the decode limits the frequency, I think adding a olo_base_pl_stage will not really help because I'd not expect synthesis tools to be able moving a pipeline stage with READY signal into the logic.

What I'd instead suggest to users is to use the _olo_base_fifo_packet_ in its original form and add an _olo_ft_ecc_encode_ and _olo_ft_ecc_decode_ on the data input/output in the surounding design (and make them registered). The user then has to care himself about the delays.

doc/ft/olo_ft_fifo_sync.md
```diff
@@ -0,0 +1,136 @@
...
+### Combinational ECC Encoder and Decoder
+
+Both codecs are instantiated with `Pipeline_g = 0`. The datapath to and from the internal FIFO is
+therefore combinational, and the _ft_ entity behaves exactly like its
+[olo_base_fifo_sync](../base/olo_base_fifo_sync.md) counterpart.
+
+The ECC decode lies between the RAM output and the output ports and is the critical path of the entity.
+Where it limits the achievable clock frequency, add an
+[olo_base_pl_stage](../base/olo_base_pl_stage.md) on the output side in the surrounding design.
+Register _Out_Data_, _Out_EccSec_ and _Out_EccDed_ in the same stage to keep them aligned.
```
obruendl: Same comment as for the fifo_packet - IMO the user would better use the base FIFO and add pipelined encode/decode around it. Do you agree?

obruendl:
@rustyqt 

I agree with the decision to make encode/decode combinatorial. It's probably the most sane trade-off. User still can implement pipelined encode/decode outside of the FIFO if they want to push frequency.

And overall the mindset between open logic is to implement things clean and simple to cover the 80% of normal cases. Special cases like optimization are not covered deliberately - because it would complicate design and interface. IMO you cleanly follow this idea with the solution chosen. That's way better than adding a bunch of obscure optimization generics or the like.

rustyqt:
@obruendl Yes, this is what i was thinking as well: When users need to push frequency they can implement their pipelined version themselves. This decision simplified the design a lot so the review should be straight forward now. Let me know if you have any additional comments on the current state.

obruendl:
@rustyqt 

It's now only the two documentation comments pending. Let me know when you tackled those, then I merge to the feature/fault-tolerant branch.

After that I think we are ready to release the open-logit FT area to the public, right? If so, I'd include it in the next release.

doc/ft/olo_ft_fifo_sync.md
rustyqt: Done. Both docs have gained a chapter named "Combinational ECC Encoder and Decoder"

rustyqt:
@obruendl 

**On the two documentation comments:** I only see one thread still open, the "Mention that this is always combinatorial (zero cycles delay)" one on `olo_ft_fifo_sync.md`. I believe that is covered by the new Combinational ECC Encoder and Decoder section, which states that both codecs run at `Pipeline_g = 0` and that the datapath to and from the internal FIFO is combinational. The equivalent section exists in `olo_ft_fifo_packet.md`. If you mean something else, could you point me at the two you mean?

**On releasing the ft area:** What is merged after this PR is self-contained and I am happy to see it released. Once released the next thing to tackle is the `olo_ft_fifo_async` which depends on two things first:

1. Fault-tolerant clock-crossing primitives. `olo_ft_cc_bits`, `olo_ft_cc_pulse` and `olo_ft_cc_reset`, TMR versions of their base counterparts. They are small and self-contained, so they would make a short PR.
2. A layering decision that touches `olo_base_fifo_async`. ECC alone is not enough for an asynchronous FIFO, the Gray-coded pointer and reset crossings need the TMR primitives above, and those sit inside the base entity. My current solution extracts the pure control logic of olo_base_fifo_async into a private entity in the same file, `olo_private_fifo_async_core`, with the RAM and the crossings exposed as ports. `olo_base_fifo_async` then adds plain CDC and `olo_ft_fifo_async` adds TMR CDC plus end to end ECC. Interface and behavior of olo_base_fifo_async are unchanged.

PR #340, opened 2026-08-23
Update olo_base_crc_append component to support CRCs wider than the bus width

rbrglez:
## Description

This pull request addresses one of the issues described in #239, specifically the limitation of the olo_base_crc_append component when the CRC width is greater than the bus width.

I have implemented most of the updated olo_base_crc_append component, but I’m struggling with the final part of the implementation.

@obruendl , could you take a look at the current implementation and provide some guidance on how to approach the remaining part?

## What I am struggling with

I added an input pipeline stage because the input stream needs to be delayed by one clock cycle, giving the CRC component enough time to calculate the CRC that will be appended to the stream.

Previously, this was not necessary because the last data beat never contained any part of the CRC. With the updated implementation, however, the last data beat can contain either the entire CRC or part of the calculated CRC.

The problem is that I now have two instances, i_crc and i_pl_in, that could potentially control the In_Ready signal. In the current implementation, i_crc controls In_Ready, while i_pl_in has no backpressure control. This can lead to data beats being lost if the Out_* interface experiences long gaps in the data stream, where neither Out_Valid nor Out_Ready is asserted.

On the other hand, if I give backpressure control to i_pl_in, the CRC append logic fails when two packets arrive back-to-back. The calculated CRC for the first packet can be lost before it is written to the output stream, while the CRC for the second packet is already being calculated.

### DIFF
--- a/sim/test_configs/olo_base.py
+++ b/sim/test_configs/olo_base.py
@@ -335,6 +335,14 @@ def add_configs(olo_tb):
     for DataWidth, CrcWidth in [(8, 8), (16, 8), (16, 16)]:
         named_config(tb, {'CrcWidth_g': CrcWidth, 'DataWidth_g': DataWidth})
 
+    ### olo_base_crc_append_be ###
+    crc_append_be_tb = 'olo_base_crc_append_be_tb'
+    tb = olo_tb.test_bench(crc_append_be_tb)  
+    CrcNames = ["CRC-8/DVB-S2", "CRC-16/DECT-X", "CRC-32/ISO-HDLC"]
+    for RandomStall in [True, False]:
+        for Crc in CrcNames:
+            named_config(tb, {'CrcName_g': Crc,'RandomStall_g' : RandomStall})
+
     ### olo_base_crc_check ###
     crc_check_tb = 'olo_base_crc_check_tb'
     tb = olo_tb.test_bench(crc_check_tb)  
--- a/src/base/vhdl/olo_base_crc_append.vhd
+++ b/src/base/vhdl/olo_base_crc_append.vhd
@@ -45,15 +45,17 @@ entity olo_base_crc_append is
         Clk              : in    std_logic;
         Rst              : in    std_logic;
         -- Input
-        In_Data          : in    std_logic_vector(DataWidth_g-1 downto 0);
-        In_Valid         : in    std_logic := '1';
         In_Ready         : out   std_logic;
+        In_Valid         : in    std_logic := '1';
         In_Last          : in    std_logic;
+        In_Be            : in    std_logic_vector(DataWidth_g/8-1 downto 0) := (others => '1');
+        In_Data          : in    std_logic_vector(DataWidth_g-1 downto 0);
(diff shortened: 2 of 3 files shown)


rbrglez:
I have updated the CRC Append so that it now passes all testbenches. I have also added a new testbench to stress-test CRC appending to the bus, specifically covering cases where the CRC is wider than the bus width.

obruendl:
I am currently on holiday in south africa and I have very limited network access.

I'll review your changes when I am back.

obruendl:
@rbrglez 

I have a conceptual question:

If I got it right, your idea for the case where CRC is wider than the data would be to configure _olo_base_crc_append_ for the CRC width and use byte-enables to only enable a certain part of the input word. Correct?

Wouldn't it potentially be easier to keep the width of _olo_base_crc_append_ still equal to the data-width and simply apply multiple beats that hold a part of the CRC? e.g. for data width = 8 and CRC width = 32 we would append 4 CRC beats.

See it as question - I might overlook something important. But I tend to tackle conceptual questions before going into details.

rbrglez:
> If I got it right, your idea for the case where CRC is wider than the data would be to configure olo_base_crc_append for the CRC width and use byte-enables to only enable a certain part of the input word. Correct?

Incorrect

> Wouldn't it potentially be easier to keep the width of olo_base_crc_append still equal to the data-width and simply apply multiple beats that hold a part of the CRC? e.g. for data width = 8 and CRC width = 32 we would append 4 CRC beats.

This is how I implemented the solution.

obruendl:
@rbrglez 

I looked quickly into it, and I think the PR does too much. It adds byte-enables and tries to append the CRC unaligned (by using the last data-bytes of the last-beat which were disabled by the BEs) ,right?

I understand that some protocol units will need this - but the ide of olo_base_crc_append is not to cover each and every case - it is to very easily protect data-streams that do not have very specific needs. I'd say it is not aimed for being used to append CRCs for wide busses and pre-defined protocols (where it is not allowed to include padding bytes in the CRC calculation).

Why: First, because this makes a simple thing very complex. And complex things tend to fail - or to not be trusted by users. Second, because it will be even harder to make the olo_crc_check() equally flexible - and I really want those to be symmetric.

When you want to implement CRC for a given protocol (e.g. UDP), I assume some protocol FSM is needed anyways at least quite often(e.g. for header/trailer generation) and this FSM can be used to append the CRC (which of course can be generated by olo_base_crc) correctly as well.

****

That being said, I see one thing we could do to achieve something similar: 
- Add BE to the olo_base_crc_append (CRC would still be word-aligned, but some bytes efore it could be disabled)
- Implement the stream-compresser unit we discussed some time back (that drops disabled bytes and compresses all data nicely).

This way a wide CRC can be appended by simply chosing olo_base_crc_append width for the CRC and keep some bytes disabled on the input data all the time and then add the stream-compresser after it. This would:
- Keep the olo_base_crc_append super simple
- Add a valuable new feature (stream compresser)
- Allow the same thing (wide CRCs and append it directly after data)

